PR #2, opened 2019-10-16
Fixes 'search offers' query by type, and other small fixes.

samhiatt:
- Fixes 'search offers' to allow query by type (changed `--type=interruptible` to `--type=bid`)
    - Adds a command line option `-b` (for "bid") which aliases `--type=bid`
    - For backwards compatibility, `--type=interruptible` is converted to `--type=bid`.
- Fixes error message caused when querying by 'dph'. 
- Fixed some duplicated and incorrect help prompts.

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -361,13 +361,14 @@ def display_table(rows, fields):
         print("  ".join(out))
 
 @parser.command(
-    argument("-t", "--type", default="on-demand", help="whether to show `interruptible` or `on-demand` offers. default: on-demand"),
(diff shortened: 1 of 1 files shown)


samhiatt:
Would you please add a LICENSE.txt file?  https://choosealicense.com/ 

I like your product and I have a bunch more work that I'd like to contribute that will hopefully make it easier to work with in python, but I'd like to ensure attribution. I prefer the MIT license since it's short and sweet. ;)

jsbcannell (approved):

PR #8, opened 2020-10-21
Remove spaces to allow for awk, sort etc.

leona:


lahwran:
looks good, thank you!

PR #11, opened 2021-10-10
add a `vast show invoices` command

bvandenbon:
This command shows the JSON which is used by the billing page of the vast.ai website.
It can only show it in Raw JSON format for now.

I added this command because the website of vast.ai no longer renders my invoices.
I assume that it just contains too much data.

Occasionally I do get a couple of PDF files, but then still my bookkeeper is pretty confused.
A better solution probably would be to make a clearer distinction between credit consumption and money transfer.
I mean, right now the invoice document is actually more like a mixture of an invoice and a credit consumption report.
And mostly the "Total" listed on top of the PDF file is confusing my bookkeeper + the fact that an invoice is several pages long and does contain a lot of neglectable numbers. (0$ amounts).

So, I wil try to give my bookkeeper 3 numbers:
- payments made by me / month.
- available credits at the start of the month.
- available credits at the end of the month.

To make these calculations I of course need the data. And that is where this command comes in.

**More importantly for this Pull Request:**
 - I tested it, and it works.


jsbcannell:
Thanks! We were actually just looking at adding this missing API endpoint, to provide other options to generate invoice data for the same reasons you list. I'm curious what tools you may be using to process the json or create a better invoice.

PR #20, opened 2022-01-20
Added the small logo file

gpropf:


gpropf:
gpu_ram bug taken care of. There is now a 1000x multiplier on that field in the searches.

PR #31, opened 2022-04-23
Copy command

gpropf:


sudohainguyen:
hey @gpropf  it requires instance password when I run copy files from local to instance. Where do I find the password?

PR #108, opened 2024-05-10
fixes: multiple uri tickets

VinTomie:
Fixes some things uri found in his automated testing primarily around ssh keys, filters, etc


VinTomie:
removing flags for transfer credit

PR #117, opened 2024-08-29
Implemented CLI Command for updating templates directly.

Marcoh572:


jsbcannell:
This was already added?

PR #120, opened 2024-09-18
feat: manage user environment variables from api

osamja:
Add CRUD CLI commands to manage user environment variables

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -1128,6 +1128,25 @@ def create__api_key(args):
         print("An unexpected error occurred:", e)
 
 @parser.command(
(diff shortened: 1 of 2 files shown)


vast.py
```diff
@@ -3599,7 +3598,83 @@ def transfer__credit(args: argparse.Namespace):
         print(r.text);
         print("failed with error {r.status_code}".format(**locals()));
 
+@parser.command(
```
jsbcannell, line 3601: commands should be sorted in the file to show up in correct sorted order

osamja: Whoops, fixed

PR #121, opened 2024-09-24
Autoscaler: add commands for easy creation of Autoscaler compatible templates

Nader-gator:
create the 4 following commands:

```
create as-tgi            Create a new autoscaler with tgi backend
create as-tgi-tmpl   Create an autoscaler compatible template for tgi
create as-cui            Create a new autoscaler with ComfyUI backend
create as-cui-tmpl   Create an autoscaler compatible template for ComfyUI
```


jsbcannell (changes requested):
Change this to a single or two command which takes the use case/template as an arg and has the naming scheme we discussed

PR #124, opened 2024-10-08
Implemented Template CRUD to work with new endpoints

Marcoh572:
@jsbcannell


karthikp32 (commented):
does this allow a user to map and unmap a template to themselves and but keep the associated template record in the templates table

Marcoh572:
Yes it does

PR #125, opened 2024-10-17
Integrate cli commands with job scheduling system

karthikp32:
Enabling these cli commands to be executed in scheduled job fashion: cloud copy, change bid,  reboot instance, execute

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -18,6 +18,7 @@ import requests
 import getpass
 import subprocess
 from subprocess import PIPE
+from croniter import croniter, CroniterBadCronError
 
 try:
     from urllib import quote_plus  # Python 2.X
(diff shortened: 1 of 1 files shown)


Nader-gator (approved):

vast.py
```diff
@@ -52,6 +53,64 @@
 class Object(object):
     pass
 
+def convert_cron_to_millis(schedule):
+    """Convert cron schedule to a time interval in milliseconds."""
+    try:
+        cron = croniter(schedule)  # Validate the cron schedule
+    except CroniterBadCronError:
+        raise argparse.ArgumentTypeError(f"Invalid cron schedule: '{schedule}'. "
+                                          "Make sure it follows the cron syntax format:\n"
```
Marcoh572, line 62: Won't this throw a error since its essentially a bunch of separate strings instead of one big one? I think you're looking for a block string with ''' abcdefg '''

vast.py
```diff
@@ -924,6 +983,12 @@ def change__bid(args: argparse.Namespace):
     r.raise_for_status()
     print("Per gpu bid price changed".format(r.json()))
 
+    if (args.schedule):
+        cli_command = "change bid"
+        api_endpoint = "/api/v0" + "/instances/bid_price/{id}/".format(id=args.id)
```
Marcoh572, line 988: why not reuse the url variable that is equivalent to api_endpoint?

vast.py
```diff
@@ -1767,6 +1877,11 @@ def execute(args):
                 if (r.status_code == 200):
                     filtered_text = r.text.replace(rj["writeable_path"], '');
                     print(filtered_text)
+                    if (args.schedule):
+                        cli_command = "execute"
+                        api_endpoint = "/api/v0" + "/instances/command/{id}/".format(id=args.ID)
```
Marcoh572, line 1882: yeah still the same url

vast.py
```diff
@@ -2179,6 +2294,11 @@ def reboot__instance(args):
         rj = r.json();
         if (rj["success"]):
             print("Rebooting instance {args.ID}.".format(**(locals())));
+            if (args.schedule):
+                cli_command = "reboot instance"
+                api_endpoint = "/api/v0" + "/instances/reboot/{id}/".format(id=args.ID)
```
Marcoh572, line 2299: and again lol- perhaps just `api_endpoint = url` if you wanna keep your descriptive variable name

vast.py
```diff
@@ -1091,17 +1157,60 @@ def cloud__copy(args: argparse.Namespace):
...
+    request_body = {
+                "start_time": start_time, 
+                "end_time": end_time, 
+                "time_interval": time_interval,  
+                "api_endpoint": api_endpoint,
+                "request_method": request_method,
+                "request_body": req_json
+            }
+                # Send a POST request
+    response = requests.post(schedule_job_url, headers=headers, json=request_body)
```
Marcoh572, line 1193: why ask for request_method if only POST requests are called?

vast.py
```diff
@@ -4597,11 +4732,16 @@ def login(args):
...
+
+
 def main():
     parser.add_argument("--url", help="server REST api url", default=server_url_default)
     parser.add_argument("--retry", help="retry limit", default=3)
     parser.add_argument("--raw", action="store_true", help="output machine-readable json")
     parser.add_argument("--explain", action="store_true", help="output verbose explanation of mapping of CLI calls to HTTPS API endpoints")
+    parser.add_argument("--schedule", help="try to schedule a command to run every x mins, hours, etc. by passing in time interval in cron syntax to --schedule option. Can also choose to have --start_time and --end_time options with valid values. For ex. --schedule \"0 */2 * * *\"")
+    parser.add_argument("--start_time", help="the start time for your scheduled job in millis since unix epoch. Default will be current time. For ex. --start_time 1728510298144", default=(time.time() * 1000))
+    parser.add_argument("--end_time", help="the end time for your scheduled job in millis since unix epoch. Default will be 7 days from now. For ex. --end_time 1729115232881", default=(time.time() * 1000 + 7 * 24 * 60 * 60 * 1000))
```
Marcoh572, line 4744: yknow you can enforce type with argparse arguments by just adding `type=int`
also if you want it so that the user can choose --schedule OR (start_time and end_time) I recommend checking out argparse's mutually exclusive groups :D up to you

vast.py
karthikp32: scheduled_jobs table is intended to store GET, POST, PUT, or DELETE requests. but call to scheduled_jobs api endpoint will be post if inserting a new record and PUT if updating

vast.py
karthikp32: url also includes hostname, port maybe, etc.

vast.py
karthikp32: seems okay to me when I tested, seems like croniter is built to handle this

vast.py
karthikp32: don't want hostname:port, etc. included

vast.py
karthikp32: don't want hostname:port, etc. included

Marcoh572 (approved):
LGTM!

vast.py
karthikp32: not mutually exclusive

PR #134, opened 2024-10-29
Fixes #133 - ssh keys should allow for file paths

### LINKED ISSUE #133: ssh keys should allow for file paths as well
Currently I can just add a ssh-key of the string ".ssh/id_rsa.pub" which is pretty useless and makes things confusing. So instead, there should be a hybrid approach that can do both file paths and ssh public keys.

kristopolous:
This makes sure the key being added looks like a public key so that people don't accidentally add a private key or some random string

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -791,6 +791,39 @@ def parse_vast_url(url_str):
 
     return (instance_id, path)
 
+def get_ssh_key(argstr):
+    ssh_key = argstr
+    # Including a path to a public key is pretty reasonable.
+    if os.path.exists(argstr):
+      with open(argstr) as f:
+        ssh_key = f.read()
+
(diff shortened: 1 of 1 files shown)


vast.py
```diff
@@ -791,6 +791,39 @@ def parse_vast_url(url_str):
 
     return (instance_id, path)
 
+def get_ssh_key(argstr):
+    ssh_key = argstr
+    # Including a path to a public key is pretty reasonable.
+    if os.path.exists(argstr):
+      with open(argstr) as f:
+        ssh_key = f.read()
```
osamja, line 799: What does reading the ssh_key here allow us to do?

osamja (commented):
One question, otherwise LGTM

osamja (approved):

PR #139, opened 2024-10-30
Fixes #138 adds tab completion

### LINKED ISSUE #138: Magical tab complete
There should be a truly magical tab completion to make lives easier.

kristopolous:
This should be completely backwards compatible and gracefully degrade on systems without the pip library installed.

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python3
+# PYTHON_ARGCOMPLETE_OK
 
 from __future__ import unicode_literals, print_function
 
 import re
 import json
 import sys
-import argparse
+import argcomplete, argparse
 import os
 import time
 from typing import Dict, List, Tuple
@@ -18,7 +19,7 @@ import requests
 import getpass
 import subprocess
 from subprocess import PIPE
-import xdg_base_dirs
+from typing import Optional
 import shutil
 import logging
 import textwrap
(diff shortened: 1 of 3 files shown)


vast.py
```diff
@@ -160,6 +171,15 @@ def load_permissions_from_file(file_path):
     with open(file_path, 'r') as file:
         return json.load(file)
 
+def complete_instance_machine(prefix=None, action=None, parser=None, parsed_args=None):
+  return show__instances(ARGS, {'internal': True, 'field': 'machine_id'})
```
osamja, line 175: One consideration is that our API rate limiting policy is quite strict at the moment.  Users pressing tab for autocomplete could unknowingly trigger enough requests to get rate limited.

kristopolous: correct. The *right* solution here is a caching layer. The XDG move allows us to instrument such an effort more properly but I made the decision that "1.0" shouldn't have a cache. Also, these are only run after the action is specified.

kristopolous: Also last night I realized there should be a readme update as part of this task

osamja (approved):
Minor feedback:
`./vast.py show <autocomplete>` does not work but `./vast.py show\ <autocomplete>` does.  Probably because of the whitespace delimiter we discussed.  

Pulled down, I think one thing that is odd is that for users who do `./vast.py <autocomplete> it shows all the commands with the whitespace escape character
image

LGTM

kristopolous:
This is correct. The *right* way to fix this is to rearrange how the fundamental parsing works.  There may be more aggressive ways to hide the slash but the invocation of say, attach ssh is 

    $ vastai "attach ssh" "instance-id" "key"

and *not*

    $ vastai "attach" "ssh" "instance-id" "key"

Which would be the actual fix. Forcing "attach ssh" to be equivalent to "attach" "ssh" as in, no escaped space, requires multiple hacks ... it's possible but imho it's not worth it.

kristopolous:
I'd of course be happy to do it after I exhaust my queue of more relevant tasks unless you think this is a deal-breaker.

PR #141, opened 2024-11-04
Further introspection work for the sdk wrapper

kristopolous:
This code gives the wrapped functions valid names, function signatures and google formatted help documentation as extracted from the argparse and docstring objects on the functions


osamja (approved):
Pulled down, tested.  LGTM

I think I like the python sdk more than the CLI

PR #152, opened 2024-11-18
Update vast.py and vast_machine_tester.py

jjziets:
This pull request includes self-test machine function to vast.py and vast_machine_tester.py.


jjziets:
improved destroy_instance_silent so that it will ensure the completed instance is deleted

PR #156, opened 2024-11-25
Delete Tempalate Redesign - CLI component

Marcoh572:
@jsbcannell I removed the ability to delete via hash id because:
1. if multiple people are using essentially the same templates with the same hash id, we wouldn't want the delete to affect more than just the one template a user is trying to delete
2. The main component of deleting templates has to do with the user_template relationship, which is based on the template_id itself, not the hash id, so it makes sense to just revert this back to only using the template id for deleting

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,5 @@ pytz
 urllib3==2.2.3
 poetry-dynamic-versioning==1.8.1
 gitpython==3.1.44
-toml=0.10.2
+toml==0.10.2
 curlify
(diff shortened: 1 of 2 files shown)


kristopolous:
Deprecating `--template-id` is a breaking change. We ok with this?

Marcoh572:
> Deprecating `--template-id` is a breaking change. We ok with this?

@kristopolous Do you mean you believe people aren't going to know how to just add the template id after the delete template command? Or are this assuming there are those who are programmatically deleting templates with that --template-id argument being explicitly typed? 

Because as far as the functionality of the command is concerned, it still sends "template_id" as a key in it's json body, which is what the endpoint needs

PR #158, opened 2024-12-01
Added few missing arguments in "create template" and updated example command

Avaray:


Avaray:
I think `--public` argument is a good idea.

jsbcannell (approved):

PR #171, opened 2025-01-13
to able to self-test deverified machines

karthik-vast-ai:


jjziets:
I also changed this in my PR. I also added auto-verify to it so if the key allows then the machine will get verified

karthik-vast-ai:
should I delete this pr then @jjziets ?

jsbcannell (approved):

jjziets:
The problem that this PR attempted to address was when the host had stopped or idle instances, aka rented them self but not running. In that case "rentable=true" will not return a result. 
By setting rentable=any the search will always return offers. Even if those are not available to rent. Such as, when one or more of the instance slots are occupied.. this will mean that if one GPU is rented, the machine cant be tested as the top offer is not available as the self test can't create the instance from an offer that is not available pr 170 has addressed this issue by setting the search to "rentable=true", "rented=any" this will return the available offer that can be rented even if the machine has an instance that is stopped under the API account that started. I have rebased my bartch and the pr 170 should address this issue if merged.

PR #181, opened 2025-03-11
Copy command - Remove sudo from rsync and add option --exclude 

coezbek:
Added --exclude to the available options of the copy command, e.g. for `--exclude .venv --exclude .git`.

I also removed the `sudo` from `rsync` and `ssh`.


vast.py
```diff
@@ -25,6 +25,7 @@
 from io import StringIO
 from typing import Optional
 import shutil
+import shlex
```
kristopolous, line 28: why is this here?

PR #182, opened 2025-03-24
Integrate job scheduling functionality

karthikp32:


vast.py
```diff
@@ -1147,6 +1183,12 @@ def change__bid(args: argparse.Namespace):
     r.raise_for_status()
     print("Per gpu bid price changed".format(r.json()))
 
+    if (args.schedule):
+        cli_command = "change bid"
+        api_endpoint = "/api/v0" + "/instances/bid_price/{id}/".format(id=args.id)
```
Marcoh572, line 1188: why separate the two strings? I'm guessing it's because you want to keep a format  where all you have to change is the second string? But the problem is, not every endpoint path starts with v0, and there's no guarantee that all future ones will either. So might as well just put the full path as one string

vast.py
```diff
@@ -6099,12 +6210,19 @@ def quote_completions(self, completions: List[str], cword_prequote: str, last_wo
...
 
+
+
 def main():
     global ARGS
     parser.add_argument("--url", help="server REST api url", default=server_url_default)
     parser.add_argument("--retry", help="retry limit", default=3)
     parser.add_argument("--raw", action="store_true", help="output machine-readable json")
     parser.add_argument("--explain", action="store_true", help="output verbose explanation of mapping of CLI calls to HTTPS API endpoints")
+    parser.add_argument("--schedule", help="try to schedule a command to run hourly, daily, or monthly. Valid values are HOURLY, DAILY, WEEKLY  For ex. --schedule DAILY")
```
Marcoh572, line 6221: use the param `choices=['HOURLY','DAILY','WEEKLY']` to force the valid choices

vast.py
```diff
@@ -6099,12 +6210,19 @@ def quote_completions(self, completions: List[str], cword_prequote: str, last_wo
...
+    parser.add_argument("--start_time", help="the start time for your scheduled job in seconds since unix epoch. Default will be current time. For ex. --start_time 1728510298", default=(time.time()))
+    parser.add_argument("--end_time", help="the end time for your scheduled job in seconds since unix epoch. Default will be 7 days from now. For ex. --end_time 1729115232", default=(time.time() + 7 * 24 * 60 * 60))
```
Marcoh572, line 6223: look into how argparse works. there are a lot of built in ways to validate the args coming in

vast.py
```diff
@@ -2200,6 +2301,11 @@ def execute(args):
                 if (r.status_code == 200):
                     filtered_text = r.text.replace(rj["writeable_path"], '');
                     print(filtered_text)
+                    if (args.schedule):
```
Marcoh572, line 2304: if someone wants to schedule a cli command for the future, doesn't that mean they don't want to do it right now? So why are we running the command still and only scheduling if it comes back as a success?

Shouldn't the flow be: if schedule -> add_scheduled_job and return. Else run the command normally

vast.py
```diff
@@ -1409,17 +1451,76 @@ def cloud__copy(args: argparse.Namespace):
     if (args.explain):
         print("request json: ")
         print(req_json)
-    
+
+    if (args.schedule):
+            cli_command = "cloud copy"
+            api_endpoint = "/api/v0" + "/commands/rclone/"
+            add_scheduled_job(args, req_json, cli_command, api_endpoint, "POST")
```
Marcoh572, line 1458: should return after this right? It shouldn't run the command now if you are scheduling it for later

vast.py
```diff
@@ -2616,7 +2722,12 @@ def reboot__instance(args):
     if (r.status_code == 200):
         rj = r.json();
         if (rj["success"]):
-            print("Rebooting instance {args.id}.".format(**(locals())));
+            print("Rebooting instance {args.ID}.".format(**(locals())));
+            if (args.schedule):
```
Marcoh572, line 2726: same as before- schedule and return, don't run the command when scheduling (or wait for the command to finish)

vast.py
```diff
@@ -1409,17 +1451,76 @@ def cloud__copy(args: argparse.Namespace):
...
+    response = requests.post(schedule_job_url, headers=headers, json=request_body)
+
+    if args.explain:
+        print("request json: ")
+        print(request_body)
+
+        # Handle the response based on the status code
+    if response.status_code == 200:
+        print(f"add_scheduled_job insert: success - Scheduling {frequency} job to {cli_command} from {start_time} to {end_time}")
+        print(response.text)
```
Marcoh572, line 1496: Is this text different from the success message you're printing above? Do we need both?

vast.py
```diff
@@ -1409,17 +1451,76 @@ def cloud__copy(args: argparse.Namespace):
...
+    elif response.status_code == 401:
+        print(f"add_scheduled_job insert: failed status_code: {response.status_code}. It could be because you aren't using a valid api_key.")
+    elif response.status_code == 422:
```
Marcoh572, line 1499: maybe add an input statement saying "existing job found. Do you want to update it (y|n)? and only update the job if they say yes. This would prevent accidental updates in case of typos or other accidents

vast.py
```diff
@@ -1409,17 +1451,76 @@ def cloud__copy(args: argparse.Namespace):
...
+        sleep(4)
+        response = update_scheduled_job(cli_command, schedule_job_url, frequency, start_time, end_time, request_body) 
+    else:
+            # print(r.text)
+        print(f"add_scheduled_job insert: failed error: {response.status_code}. Response body: {response.text}")        
+
+def update_scheduled_job(cli_command, schedule_job_url, frequency, start_time, end_time, request_body):
+    response = requests.put(schedule_job_url, headers=headers, json=request_body)
```
Marcoh572, line 1507: before this maybe add a print statement saying "Updating Existing Job..." or something to provide some indication that that is happening

vast.py
karthik-vast-ai: yeah I made the assumption that they want to run now and schedule for the future. don't know if that's the best assumption to make

vast.py
karthikp32: done

karthikp32: done

vast.py
karthikp32: done

vast.py
karthikp32: done

vast.py
karthik-vast-ai: done

vast.py
karthik-vast-ai: only scheduling job instead of doing both now

vast.py
karthik-vast-ai: only scheduling job instead of doing both now

vast.py
```diff
@@ -2598,6 +2741,11 @@ def prepay__instance(args):
 
 @parser.command(
     argument("id", help="id of instance to reboot", type=int),
+    argument("--schedule", choices=["HOURLY", "DAILY", "WEEKLY"], help="try to schedule a command to run hourly, daily, or monthly. Valid values are HOURLY, DAILY, WEEKLY  For ex. --schedule DAILY"),
+    argument("--start_time", type=validate_seconds, help="the start time for your scheduled job in integer seconds since unix epoch. Default will be current time. For ex. --start_time 1728510298", default=(int(time.time()))),
+    argument("--end_time", type=validate_seconds, help="the end time for your scheduled job in integer seconds since unix epoch. Default will be 7 days from now. For ex. --end_time 1729115232", default=(int(time.time() + 7 * 24 * 60 * 60))),
+    argument("--day", help="day of the week you want scheduled job to run on. You can set day to None if you want the job to run everyday. Valid values are 0-6, 0=Sunday, 1=Monday, etc. Default will be 0. For ex. --day 0", default=0),
```
Marcoh572, line 2747: if you are validating the others you could at least validate day as well. its just `choices=range(7)`

Same for hour but `choices=range(24)`

Apart from that, good use of argparse

vast.py
```diff
@@ -2155,6 +2286,11 @@ def detach__ssh(args):
...
+    argument("--hour", help="hour of the day you want scheduled job to run on. You can set day and hour to None if you want the job to run every hour. Valid values are 0-23, 0=12am UTC, 1=1am UTC, etc. Default will be 0. For ex. --hour 16", default=0),
```
Marcoh572, line 2293: same validation as last comment for day and hour ints

vast.py
Marcoh572: why sleep(4)? kinda long isnt it?

vast.py
```diff
@@ -1346,6 +1392,11 @@ def vm__copy(args: argparse.Namespace):
...
     argument("--update", help="skip files that are newer on the destination", action="store_true"),
     argument("--delete-excluded", help="delete files on dest excluded from transfer", action="store_true"),
+    argument("--schedule", choices=["HOURLY", "DAILY", "WEEKLY"], help="try to schedule a command to run hourly, daily, or monthly. Valid values are HOURLY, DAILY, WEEKLY  For ex. --schedule DAILY"),
+    argument("--start_time", type=validate_seconds, help="the start time for your scheduled job in integer seconds since unix epoch. Default will be current time. For ex. --start_time 1728510298", default=(int(time.time()))),
+    argument("--end_time", type=validate_seconds, help="the end time for your scheduled job in integer seconds since unix epoch. Default will be 7 days from now. For ex. --end_time 1729115232", default=(int(time.time() + 7 * 24 * 60 * 60))),
+    argument("--day", help="day of the week you want scheduled job to run on. You can set day to None if you want the job to run everyday. Valid values are 0-6, 0=Sunday, 1=Monday, etc. Default will be 0. For ex. --day 0", default=0),
```
Marcoh572, line 1398: day/hour validation

vast.py
karthik-vast-ai: changed to use cron parse function since "*" is an option now

vast.py
```diff
@@ -1420,6 +1477,69 @@ def cloud__copy(args: argparse.Namespace):
         print("failed with error {r.status_code}".format(**locals()));
 
 
+def add_scheduled_job(args, req_json, cli_command, api_endpoint, request_method, instance_id=None):
+    if args.start_time >= args.end_time:
+        raise ValueError("--start_time must be less than --end_time.")
```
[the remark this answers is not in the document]
karthik-vast-ai: reusing convert_dates_to_timestamps function now 
keeping that because I'm not validating start_date <= end_date

vast.py
karthik-vast-ai: changed to use cron parse function since "*" is an option now

vast.py
```diff
@@ -1124,6 +1157,11 @@ def cancel__sync(args: argparse.Namespace):
...
+    argument("--schedule", choices=["HOURLY", "DAILY", "WEEKLY"], help="try to schedule a command to run hourly, daily, or monthly. Valid values are HOURLY, DAILY, WEEKLY  For ex. --schedule DAILY"),
+    argument("--start_time", type=validate_seconds, help="the start time for your scheduled job in integer seconds since unix epoch. Default will be current time. For ex. --start_time 1728510298", default=(int(time.time()))),
+    argument("--end_time", type=validate_seconds, help="the end time for your scheduled job in integer seconds since unix epoch. Default will be 7 days from now. For ex. --end_time 1729115232", default=(int(time.time() + 7 * 24 * 60 * 60))),
+    argument("--day", help="day of the week you want scheduled job to run on. You can set day to None if you want the job to run everyday. Valid values are 0-6, 0=Sunday, 1=Monday, etc. Default will be 0. For ex. --day 0", default=0),
+    argument("--hour", help="hour of the day you want scheduled job to run on. You can set day and hour to None if you want the job to run every hour. Valid values are 0-23, 0=12am UTC, 1=1am UTC, etc. Default will be 0. For ex. --hour 16", default=0),
```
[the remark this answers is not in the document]
karthik-vast-ai: changed to use cron parse function since "*" is an option now

vast.py
```diff
@@ -1158,14 +1158,46 @@ def default_start_date():
...
+    raise argparse.ArgumentTypeError("Day must be 0-6 (0=Sunday) or '*' for every day.")
+
+def parse_hour_cron_style(value):
+    """
+    Accepts an integer string 0-23 or '*' to indicate 'Every hour'.
+    Returns 0-23 as int, or None if '*'.
+    """
+    val = str(value).strip()
+    if val == "*":
+        return "None"
```
Marcoh572, line 1184: is this supposed to be a string?

karthik-vast-ai: fixed

Marcoh572 (approved):
LGTM

skiddar (commented):
I think we should tell users what the bandwidth price and maybe make them confirm.  Also, if this fails, where does the notification of the failure go?

karthik-vast-ai:
changed default end date to contract's end date for cloud copy scheduled jobs

PR #187, opened 2025-04-04
automating generation of ssh key pair

karthik-vast-ai:


stackanthony (approved):

vast.py
```diff
@@ -1464,19 +1465,63 @@ def create__env_var(args):
...
+        print("Automation is only supported on Linux/MacOS. Please provide your SSH public key manually.")
+        return None
+
+    key_path = os.path.expanduser("~/.ssh/id_rsa")
+    pub_key_path = key_path + ".pub"
+    
+    # Generate the key pair if the public key does not exist.
+    if not os.path.exists(pub_key_path):
+        print("Generating a new RSA SSH key pair...")
+        subprocess.run(["ssh-keygen", "-t", "rsa", "-f", key_path, "-q", "-N", ""], check=True)
```
skiddar, line 1483: Why isn't this in a try catch block?

vast.py
```diff
@@ -1464,19 +1465,63 @@ def create__env_var(args):
...
+        print("SSH key pair already exists, using the existing key.")
+    
+    # Load the key into the SSH agent.
+    try:
+        subprocess.run(["ssh-add", key_path], check=True)
+    except subprocess.CalledProcessError:
+        print("Unable to add SSH key to the agent. Make sure the ssh-agent is running.")
+    
+    # List loaded keys to confirm.
+    subprocess.run(["ssh-add", "-l"], check=True)
```
skiddar, line 1494: Try/catch block?

skiddar (commented):
Looks ok, but ssh-keygen may not be installed on all distros.

vast.py
kristopolous: don't do rsa. We don't need compatibility with 2010

PR #188, opened 2025-04-04
feat/prompt-user-for-update

stackanthony:
This PR addresses the need to prompt users to update their CLI if it's out of date compared to the stable version (PyPi latest version). This works for the Pypi and git distribution. If we detect you're running the python file directly (git distribution), we prompt an update, and if they confirm, we checkout the tagged commit for them. If they're using the pip distribution, we just simply update the package for them. 

Also, there's now a `--version` flag that you can pass in to check what version of the CLI you're using. 

The workflow is setup so that whenever there's a tagged commit that matches semver pattern, it'll trigger a build and publish off to Pypi.

Before we merge this into master, we'll have to replace the all references of `vast-cli-fork` to `vastai` to be consistent with package naming previously.

(Sorry for the amount of commits, I had to trigger a lot of different gh actions to test this, I'll make sure we squash this into one semantic commit)


.github/workflows/test-python-publish.yml
```diff
@@ -0,0 +1,31 @@
+name: Test PyPi Publish
```
liam-at-vast, line 1: what's this for vs python-publish?

utils/pypi_api.py
```diff
@@ -0,0 +1,24 @@
...
+
+BASE_PATH = "https://pypi.org"
+def get_project_data(project_name: str) -> dict[str, dict[str, str]]:
+    url = BASE_PATH + f"/pypi/{project_name}/json"
+    response = requests.get(url, headers={
+        "Accept": "application/json"
+    })
+
+    if response.status_code != 200:
+        raise Exception("Non-200 Response Code")
```
liam-at-vast, line 11: probably want to give the reason, can you do response.raise_for_status()? might still need to throw because it doesn't throw on 200 <= error_code <= 399 I think, but

utils/version_checker.py
```diff
@@ -0,0 +1,114 @@
...
+                update_command,
+                shell=True,
+                check=True,
+                text=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+
+            print("Update completed successfully!")
+            print("Please restart the CLI manually to use the new version.")
```
liam-at-vast, line 114: Do they actually need to restart anything or is it installed in place?

vast.py
```diff
@@ -6159,6 +6164,9 @@ def main():
     if args.api_key:
         headers["Authorization"] = "Bearer " + args.api_key
 
+    # WARNING - will throw error if it can't install update
+    check_for_update()
```
liam-at-vast, line 6168: maybe wrap in a try/except? I think it would be frustrating if the function errors for some reason and they are prevented from using the CLI

.github/workflows/test-python-publish.yml
stackanthony: This was just for uploading to test-pypi (they have some test repo) for initial testing. I'll just comment it out for reference but it's not useful at this point

utils/version_checker.py
stackanthony: For the git distribution yeah, that will check out the tagged commit that's associated with the pypi version. For the pypi package, probably can be done in place, I can look into that

vast.py
stackanthony: got it

utils/version_checker.py
stackanthony: So looking into it, you have to reload the process (trash old and make new process) using an os method, which needs to be managed differently between Windows, and unix based systems. I think for now let's do this manually, then if we get some customer feedback on it in the future we can change it. It's also possible via the git version since we're just loading a new process up

utils/pypi_api.py
stackanthony: Got it

liam-at-vast (approved):

Marcoh572 (approved):
looks great!

PR #203, opened 2025-04-22
Take snapshot 2

karthik-vast-ai:


edgaratvast (approved):

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
         print(r.text);
         print("failed with error {r.status_code}".format(**locals()));
 
+@parser.command(
+    argument("instance_id",      help="instance_id of the container instance to snapshot",      type=str),
+    argument("--container_registry", help="Container registry to push the snapshot to. Default will be docker.io", type=str, default="docker.io"),
+    argument("--personal_repo",    help="Docker repo to push the snapshot to",     type=str),
```
skiddar, line 1543: Don't like the use of "personal_repo" here.  Since we are going after corporate accounts.

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
...
+    argument("--docker_login_user",help="Username for container registry with personal repo",     type=str),
```
skiddar, line 1544: "personal" also used here.

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
...
+    argument("--docker_login_pass",help="Password or token for container registry with personal repo",     type=str),
```
skiddar, line 1545: and here...

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
...
+    argument("--pause",            help="Pause container's processes being executed by the CPU before docker commit (true/false). Default will be true", type=str, default="true"),
```
skiddar, line 1546: Was --pause and not pause both tested to see if they can mess up jobs?

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
...
+    usage="vastai take snapshot INSTANCE_ID "
+          "--personal_repo REPO --docker_login_user USER --docker_login_pass PASS"
```
skiddar, line 1548: "personal" again

vast.py
```diff
@@ -1537,6 +1537,72 @@ def cloud__copy(args: argparse.Namespace):
...
+
+    # POST to the snapshot endpoint
+    r = http_post(args, url, headers=headers, json=req_json)
+    r.raise_for_status()
+
+    if r.status_code == 200:
+        data = r.json()
+        if data.get("success"):
+            print(f"Snapshot request sent successfully. Please check your repo {repo} in container registry {container_registry} in 5-10 mins")
+        else:
```
skiddar, line 1600: How do you know it will take 5-10 minutes?  Isn't size dependent?

skiddar (commented):
Are there any more detailed logs?  Where do they go?  
If the CPU isn't paused, does the backup run in the background?
Each backup just adds to the number of backups?  There is no, "Daily backups for XX days, or something like that, that is typical of backup services?"
If you don't have a restore function, how did you test this?
What happens if it is used on a VM container?

vast.py
karthik-vast-ai: done

vast.py
karthik-vast-ai: changed

vast.py
karthik-vast-ai: changed

vast.py
karthik-vast-ai: that was just a buffer in case it takes longer
added that it can take longer depending on the size

vast.py
karthik-vast-ai: ran simple pytorch script with a training loop with 100000 iterations and the training loop/script seemed to continue just fine when snapshot was taken with or without pausing

karthik-vast-ai:
@skiddar 
"Are there any more detailed logs? Where do they go?" - this is a V0 implementation. we can work on things like logging in a future implementation.
"If the CPU isn't paused, does the backup run in the background" - snapshot is taken regardless of whether processes being executed by CPU are paused
"Each backup just adds to the number of backups? There is no, "Daily backups for XX days, or something like that, that is typical of backup services?" - if we approve my other pr, I can add add_scheduled_job() function to this function too and we can schedule jobs to do this too
"If you don't have a restore function, how did you test this?" - I just created instance from image
"What happens if it is used on a VM container?" - this can't be used on VM instances. As far as the VM container that launches VM instances, I think it can take a snapshot of that too if it's just a docker container

karthik-vast-ai:
will prevent taking snapshots when VMs are used as part of restore snapshot API endpoint

PR #204, opened 2025-04-23
Volume fixes, defjob fixes

liam-at-vast:
don't send defaults to webserver, because this prevents us from updating fields without setting other fields to the default, i.e, if I just want to update price I can call `vastai list volume 179 -p .4` and it will update my price to .4 without changing allocation.


jsbcannell:
fix conflict

PR #211, opened 2025-05-14
removed dependency version pinning

sitmo:
In the dependencies section in pyproject.toml, changes pinned version like

"argcomplete==3.5.1",
  
to ranges like 

"argcomplete (>=3.5.1,<4.0.0)",

to improve interoperability and reduce version conflict when useing the vast.ai CLI in other environments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,20 +21,21 @@ requires-python = ">3.9.1, <4.0"
 license = { text = "MIT" }
 dynamic = ["version"]
 dependencies = [
-  "xdg",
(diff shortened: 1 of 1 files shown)


kristopolous:
the ranges are ok, but there's too much here. we want to keep some of them open. It should be a more minimal pr.

stackanthony:
To add onto this, make sure you run `poetry lock` after you've finished Chris' requested changes. This is essential to keep the state of our lockfile fresh in the project.

PR #212, opened 2025-05-15
adding --curl to search_templates

kristopolous:


stackanthony:
At some point, there should be a task for replacing the deprecated request methods with the new http wrappers you created. The mismatch between the flags that should be available to all commands will cause confusion for users

stackanthony (approved):

kristopolous:
i'm fine with it but I'm also fine with just doing it piecemeal. We'll get them all eventually

PR #219, opened 2025-06-10
Update vast CLI to use proper autoscaling names [AUTO-303]

nakinnubis:
Change the naming of autogroup to workergroup. Also modify the requirement.txt to match the pyproject.toml
[Working Demo. Please be sure to test as well](https://github.com/user-attachments/assets/59b7a322-ef6b-43a3-88bd-96e6f4429a16)

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -507,7 +507,7 @@ def translate_null_strings_to_blanks(d: Dict) -> Dict:
     new_d = {k: translate_nulls(v) for k, v in d.items()}
     return new_d
 
-    #req_url = apiurl(args, "/instances", {"owner": "me"});
+    #req_url = apiurl(args, "/instances", {"owner": "me"})
 
 
 def apiurl(args: argparse.Namespace, subpath: str, query_args: Dict = None) -> str:
@@ -1006,7 +1006,7 @@ def parse_query(query_str: str, res: Dict = None, fields = {}, field_alias = {},
             value = numeric_version(value)
 
         if not field in fields:
-            print("Warning: Unrecognized field: {}, see list of recognized fields.".format(field), file=sys.stderr);
+            print("Warning: Unrecognized field: {}, see list of recognized fields.".format(field), file=sys.stderr)
         if not op_name:
             raise ValueError("Unknown operator. Did you forget to quote your query? " + repr(op).strip("u"))
         if op_name in ["in", "notin"]:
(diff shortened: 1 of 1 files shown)


ColterDowning:
Looks good to me, but where do we update the '--help' command? I don't see it here?

nakinnubis:
> Looks good to me, but where do we update the '--help' command? I don't see it here?

I updated the helper text as well. you can download and validate it.

nakinnubis:
> Looks good to me, but where do we update the '--help' command? I don't see it here?

Thanks for noting that!.

Scott-Laytart:
I just have 2 suggestions for help files. looks like something got missed in the help for `delete workergroup`:

`Note that deleteing an workergroup group doesn't automatically destroy all the instances that are associated with your workergroup group.`

Should fix both instances where workergroups are called` workergroup group`. 

And then i think maybe we should be consistent and make sure we aren't referring to `workergroup` as `worker group` in some places and not others. Should maybe just refer to them as workergroup everywhere in the cli since that is consistent with the command structure?

nakinnubis:
> I just have 2 suggestions for help files. looks like something got missed in the help for `delete workergroup`:
> 
> `Note that deleteing an workergroup group doesn't automatically destroy all the instances that are associated with your workergroup group.`
> 
> Should fix both instances where workergroups are called` workergroup group`.
> 
> And then i think maybe we should be consistent and make sure we aren't referring to `workergroup` as `worker group` in some places and not others. Should maybe just refer to them as workergroup everywhere in the cli since that is consistent with the command structure?

I have updated it. check if it's as described. Thanks!

vast.py
```diff
@@ -1891,23 +1891,23 @@ def create__ssh_key(args):
...
+    argument("--test_workers",help="number of workers to create to get an performance estimate for while initializing workergroup (default 3)", type=int, default=3),
     argument("--gpu_ram",     help="estimated GPU RAM req  (independent of search string)", type=float),
     argument("--search_params", help="search param string for search offers    ex: \"gpu_ram>=23 num_gpus=2 gpu_name=RTX_4090 inet_down>200 direct_port_count>2 disk_space>=64\"", type=str),
     argument("--min_load", help="[NOTE: this field isn't currently used at the autojob level] minimum floor load in perf units/s  (token/s for LLms)", type=float),
     argument("--target_util", help="[NOTE: this field isn't currently used at the autojob level] target capacity utilization (fraction, max 1.0, default 0.9)", type=float),
     argument("--cold_mult",   help="[NOTE: this field isn't currently used at the autojob level]cold/stopped instance capacity target as multiple of hot capacity target (default 2.0)", type=float),
-    usage="vastai autogroup create [OPTIONS]",
-    help="Create a new autoscale group",
+    usage="vastai workergroup create [OPTIONS]",
```
Scott-Laytart, line 1902: I believe this should be:

usage="vastai create workergroup [OPTIONS]",

Looks like this was an existing bug.

PR #221, opened 2025-06-12
add functionality for adding or linking a volume to a created instance

Marcoh572:
adding args and validation for creating or linking volumes with the new instance

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -2004,6 +2004,30 @@ def get_runtype(args):
 
     return runtype
 
+def validate_volume_params(args):
+    if args.volume_size and not args.create_volume:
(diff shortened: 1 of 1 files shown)


sammy-vastai:
@Marcoh572 I believe this is breaking the self-verification test

sammy@VAST-LAX-ZB24:~/workspace/vast-cli$ ./vast.py self-test machine 38348
Machine ID 38348 meets all the requirements.
Starting test with vastai/test:self-test-cu128
Error creating instance: 'Namespace' object has no attribute 'create_volume'
{'success': False, 'reason': 'Failed to create instance. Check the docker configuration. Use the self-test machine function in vast cli '}

PR #235, opened 2025-07-18
Update vast.py - added ssh keygen

Scott-Laytart:
Added ssh keygen

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -1922,20 +1922,132 @@ def create__env_var(args):
         print(f"Failed to create environment variable: {result.get('msg', 'Unknown error')}")
 
 @parser.command(
-    argument("ssh_key", help="add the public key of your ssh key to your account (form the .pub file)", type=str),
-    usage="vastai create ssh-key ssh_key",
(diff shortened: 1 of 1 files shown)


vast.py
```diff
@@ -1922,20 +1922,132 @@ def create__env_var(args):
...
+    try:
+        ssh_dir.mkdir(mode=0o700, exist_ok=True)
+    except OSError as e:
+        print(f"Error creating .ssh directory: {e}", file=sys.stderr)
+        sys.exit(1)
+    
+    # Check if any part of the key pair already exists
+    if private_key_path.exists() or public_key_path.exists():
+        print(f"SSH key pair 'id_rsa' already exists in {ssh_dir}")
+        response = input("Would you like to generate a new key pair? THIS WILL REPLACE YOUR EXISTING KEY. [y/N]: ").lower()
```
Nader-gator, line 1994: we shouldn't delete existing keys, someone might accidentally hit yes. instead rename them to `{filename}.backup-{timestamp}` and then place the key in their old place

edgaratvast (approved):

Nader-gator (approved):

PR #255, opened 2025-08-08
Fix wrong raw output during `vastai attach ssh --raw` cmd call

borisermakof:
Hey folks!

This PR fixes two issues:
- Incorrect attach ssh command description.
- Missing proper JSON output when the `--raw` parameter is used.

While working with the vastai CLI, I noticed that my code could not parse JSON when running:
```
vastai attach ssh <params> --raw
```
This is because the command returned plain text, which is not useful for programmatic parsing.

Example before the fix:
```
❯ python vast.py attach ssh <id> <path>/keys/vastai_key.pub --raw  

"{'success': True, 'msg': 'SSH key added to instance.', 'key': '{"id": <id>, "user_id": null, "default": null, "public_key": "<keydata>", "private_key": null, "created_at": <date>, "deleted_at": null}'}  
null
```
The output is a string containing a plain text, not valid JSON.
There’s also an extra null line at the end. This forces me to do additional parsing/replacing before I can run `json.loads(output)`.

Example after the fix:
```
❯ python vast.py attach ssh <id> <path>keys/vastai_key.pub --raw  

{
  "key": "{\"id\": <id>, \"user_id\": null, \"default\": null, \"public_key\": \"<key_data>\", \"private_key\": null, \"created_at\": 1754655150.416915, \"deleted_at\": null}",
  "msg": "SSH key added to instance.",
  "success": true
}
```
Now the output is valid JSON and can be directly passed to `json.loads()` without additional processing.


vastzuby (approved):
looks good, thanks for contributing :)

PR #259, opened 2025-08-11
add cluster id filter

stackanthony:


vastzuby:
cluster_id is already added

PR #267, opened 2025-09-02
Copy test verififcation

looplater:


robballantyne (approved):
Working as intended.  Looks good to me @looplater

PR #273, opened 2025-10-01
add examples for body params

zubyjaved:
- add examples/defaults for body params

https://vastai.atlassian.net/browse/CLN-1173


zubyjaved:
not needed

PR #365, opened 2026-04-18
feat(CLN-3107): implement accept price increase command and API integration

abiola-vastai:
CLI component of price accept increase.

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -3679,81 +3679,6 @@ def logs(args):
 
 
 @parser.command(
-    argument("ids", help="instance IDs to accept (one or more). Omit when --host is used.",
-             type=int, nargs='*'),
-    argument("--host", dest="host_id",
-             help="accept every pending price-increase from this host (mutually exclusive with IDs)",
-             type=int),
-    usage="vastai accept price-increase ID [ID ...] | --host HOST_ID",
-    help="Accept a pending host price increase on one or more instances",
-    epilog=deindent("""
-        When a host raises the rental price, the auto-extend mechanism stops
-        renewing your affected contracts and an email notification is sent.
-        This command accepts the new price and lets the affected instances
-        keep auto-extending at the updated rate.
-
-        Batch accepts are capped at 64 instance IDs per call (server-side limit).
-
-        Examples:
-            vastai accept price-increase 123
-            vastai accept price-increase 123 456 789
-            vastai accept price-increase --host 12345
-    """),
-)
-def accept__price_increase(args):
-    """
-    :param argparse.Namespace args: should supply all the command-line options
-    :rtype:
-    """
-    ids = list(getattr(args, "ids", None) or [])
-    host_id = getattr(args, "host_id", None)
-
-    if bool(ids) == (host_id is not None):
-        msg = ("Provide either instance IDs or --host, not both."
-               if ids else
-               "Provide either one or more instance IDs or --host.")
-        print(msg, file=sys.stderr)
-        sys.exit(1)
-
-    if len(ids) > 64:
-        print(f"Too many instance IDs: {len(ids)} > 64 (server limit).", file=sys.stderr)
-        sys.exit(1)
-
-    if len(ids) == 1 and host_id is None:
-        url = apiurl(args, "/instances/{id}/accept-price-increase/".format(id=ids[0]))
-        json_blob = {}
-    else:
-        url = apiurl(args, "/instances/accept-price-increase/")
-        json_blob = {}
-        if ids:
-            json_blob["instance_ids"] = ids
-        if host_id is not None:
-            json_blob["host_id"] = host_id
-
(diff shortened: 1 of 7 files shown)


vastai/cli/commands/instances.py
```diff
@@ -634,6 +634,92 @@ def change__bid(args):
...
+               if ids else
+               "Provide either one or more instance IDs or --host.")
+        print(msg, file=sys.stderr)
+        sys.exit(1)
+
+    if len(ids) > 64:
+        print(f"Too many instance IDs: {len(ids)} > 64 (server limit).", file=sys.stderr)
+        sys.exit(1)
+
+    if args.explain:
```
vastzuby, line 686: claude suggestion. not sure if needed 

**`--explain` output is misleading for single-id**

`vastai/cli/commands/instances.py:683-685`

For `vastai accept price-increase 123 --explain`, the block prints:

```
request json:
{'id': 123}
```

But the actual request is `PUT /instances/123/accept-price-increase/` with body `{}` — the id is path-encoded, never in the body. The synthetic `{"id": ids[0]}` here is never sent over the wire.

Concrete confusion: anyone copying that body into `curl` to reproduce won't match what the CLI does; anyone debugging a server-side read of `id` will chase the wrong source.

Suggest printing what's actually sent, and including the URL so the path-vs-body route distinction is visible:

```python
if args.explain:
    if len(ids) == 1:
        print(f"PUT /instances/{ids[0]}/accept-price-increase/")
        print("request json: {}")
    else:
        payload = {}
        if ids:
            payload["instance_ids"] = ids
        if host_id is not None:
            payload["host_id"] = host_id
        print("PUT /instances/accept-price-increase/")
        print(f"request json: {payload}")
```

vastzuby:
@abiola-vastai  also `vast.py` is deprecated. i wouldnt worry about updating it. we're just keeping the old commands in there for backwards compatibility.

vastai/cli/commands/instances.py
abiola-vastai: I will take a look thanks

vast.py
```diff
@@ -8188,7 +8268,7 @@ def list__machine(args):
     argument("-d", "--price_inetd", help="price for internet download bandwidth in $/GB", type=float),
     argument("-b", "--price_min_bid", help="per gpu minimum bid price floor in $/hour", type=float),
     argument("-r", "--discount_rate", help="Max long term prepay discount rate fraction, default: 0.4 ", type=float),
-    argument("-m", "--min_chunk", help="minimum amount of gpus", type=int),
+    argument("-m", "--min_chunk", help="minimum amount of gpus (default: 1)", type=int, default=1),
```
abiola-vastai: Thanks I will take a look

vastai/cli/commands/machines.py
```diff
@@ -228,7 +233,7 @@ def list__machine(args):
     argument("-d", "--price_inetd", help="price for internet download bandwidth in $/GB", type=float),
     argument("-b", "--price_min_bid", help="per gpu minimum bid price floor in $/hour", type=float),
     argument("-r", "--discount_rate", help="Max long term prepay discount rate fraction, default: 0.4 ", type=float),
-    argument("-m", "--min_chunk", help="minimum amount of gpus", type=int),
+    argument("-m", "--min_chunk", help="minimum amount of gpus (default: 1)", type=int, default=1),
```
vastzuby, line 236: nit: the help text change is fine but i'd remove the last part (default=1)

the backend already set the default to 1

vastai/cli/commands/machines.py
```diff
@@ -199,7 +199,7 @@ def list_machine_impl(args, id):
...
```
vastzuby, line 202: nit: the help text change is fine but i'd remove the last part (default=1)

the backend already set the default to 1

vastai/cli/commands/machines.py
vastzuby: @abiola-vastai  actually keep it in. its a good change 

without it, `vastai list machine` without -m 400's

1. user calls `list machine` without -m 
2. argparse returns None
3. we send `{min_chunk: null}` to backend
4. backend fails at `int(params.get("min_chunk, 1))`. which looks like it would work but `.get()` only defaults to 1 if they key is missing. in this case its explicity set to null so raises error

vastai/cli/commands/machines.py
vastzuby: @abiola-vastai actually keep it in. its a good change

without it, vastai list machine without -m 400's

    user calls list machine without -m
    argparse returns None
    we send {min_chunk: null} to backend
    backend fails at int(params.get("min_chunk, 1)). which looks like it would work but .get() only defaults to 1 if they key is missing. in this case its explicity set to null so raises error

vastzuby (approved):

PR #374, opened 2026-04-23
run CLI tests on all 3 OSes via matrix, drop OS smoketest

vastzuby:
## Summary
  - Adds OS matrix (ubuntu/windows/macos) to unit-and-integration so new pytests
    run on every OS automatically.
  - Adds a small install-smoke matrix job (`pip install . && vastai --help`) to
    catch packaging regressions unit tests miss.
  - Ports the api-key write round-trip and xdg-ImportError fallback from the
    old inline smoketest into pytest under tests/cli/ and tests/sdk/.
  - Deletes os-smoketest.yml (its purpose is now covered).
  - Uses snok/install-poetry@v1 so Poetry installs on Windows (no python3 alias).

  ## Test plan
  - [ ] CI green on ubuntu-latest, windows-latest, macos-latest for both jobs
  - [ ] First Windows run may surface ubuntu-only assumptions in existing tests
        — those are real bugs to fix, not a reason to back out the matrix
  - [ ] Serverless-testing intentionally stays ubuntu-only (separate scope)


.github/workflows/vast-sdk-testing.yml
```diff
@@ -49,28 +49,79 @@ jobs:
...
     if: needs.detect-changes.outputs.sdk == 'true'
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
+        # Avoid fetch-depth: 0 — the repo has a remote ref with a `"` in its
```
LucasArmandVast: Possible to remove this commit from git history? Or at least rename?

vastzuby: done

LucasArmandVast (approved):

PR #376, opened 2026-04-24
fix: clean up entered queue at that is no longer needed

madisonvast:
## Summary
- Remove the `entered_queue_at` field from the `RequestMetrics` dataclass
- Stop sending `entered_queue_at` in the delete_requests payload to the autoscaler
- Drop the two `entered_queue_at=time.time()` kwargs in the request constructor calls

The field is no longer read by the autoscaler. Queue time metrics are computed
from `created_at` (set autoscaler-side when the request first arrives) and
`work_started_at`, so `entered_queue_at` was redundant on the wire and unused
downstream.

## Test plan
- [ ] Verify pyworker still serializes delete_requests payloads correctly
- [ ] Verify autoscaler accepts the trimmed payload without errors
- [ ] Confirm queue time metrics in Prometheus continue to populate

### DIFF
--- a/vastai/serverless/server/lib/backend.py
+++ b/vastai/serverless/server/lib/backend.py
@@ -308,7 +308,6 @@ class Backend:
             workload=auth_data.get("cost"),
             status="SessionActive",
             is_session=True,
-            entered_queue_at=time.time(),
         )
 
         async with self._sessions_lock:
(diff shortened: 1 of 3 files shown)


vastzuby (approved):

LucasArmandVast:
See my comment on your other PR, but I think this is still useful information that we want to report/keep. No pyworker changes needed

PR #381, opened 2026-04-28
add serverless benchmarking command

vastzuby:


vastai/api/endpoints.py
```diff
@@ -347,6 +347,31 @@ def get_wrkgrp_logs(client, id, level=1, tail=None):
...
+    Args:
+        client: VastClient instance.
+        id (int): ID of endpoint group whose workers to list.
+
+    Returns:
+        list | dict: Worker data from the API response.
+    """
+    base = _get_logs_base_url(client)
+    url = base + "/get_endpoint_workers/"
+    json_blob = {"id": id, "api_key": client.api_key}
```
LucasArmandVast: Is it required to put API key in JSON body here?

vastai/cli/commands/benchmark.py
```diff
@@ -0,0 +1,774 @@
...
+
+def _count_matching_offers(client, *, gpu_name, num_gpus, extra_filters):
+    """Return how many verified+rentable offers match the template's filters
+    plus ``gpu_name`` and ``num_gpus``. ``limit=1`` because we only care
+    whether any exist.
+    """
+    query = dict(extra_filters or {})
+    query["gpu_name"] = {"eq": gpu_name}
+    query["num_gpus"] = {"eq": num_gpus}
+    offers = _api_with_retry(offers_api.search_offers,
```
LucasArmandVast: This seems like something we can just replace with the SDK itself

vastai/api/endpoints.py
vastzuby: every other endpoint does this too. more for consistency. 

we should remove all of them though in a future PR and just keep it in the bearer header

vastai/cli/commands/benchmark.py
vastzuby: yeah ill switch to using the client for these and address the rest of the issues you brought up. thanks 👍

vastzuby:
@LucasArmandVast  addressed your feedback 

1. use existing sdk functionality 
2. fix ctrl + c hanging issue, now prompts user to let the endpt finish deleting before they ctrl + c a second time 
3. rename endpoint to show gpu name + count
4. increase timeout from 30m -> 60m. if no measured perf by then, aborts the command 

i didnt switch to async right now since the AsyncClient doesnt have any serverless functions. So i could add them to the async client but i think for MVP, using ThreadPoolExecutor is ok

vastai/cli/commands/benchmark.py
```diff
@@ -0,0 +1,769 @@
...
+            return (gpu_name, "error", None,
+                    f"create_endpoint returned no id: {ep_resp!r}", None)
+        active_endpoints.add(endpoint_id)
+        _update_row(class_states, gpu_name, endpoint_id=endpoint_id)
+
+        wg_kwargs = dict(
+            endpoint_id=endpoint_id,
+            endpoint_name=endpoint_name,
+            search_params=search,
+            test_workers=1,
```
LucasArmandVast: unused parameter, just remove

vastzuby: done

vastzuby:
@LucasArmandVast 

1. fix regression not showing why the benchmark failed (usually from template config)
2. deduplicate same gpus passed in arg
3. improve help text
4. fix table not showing same gpu with different chunks
5. allow no spaces between num gpus and name ("8xRTX_5090" and "8x RTX 5090" both work)

LucasArmandVast (approved):

PR #384, opened 2026-04-29
fix: owner: me query arg

### LINKED ISSUE #382: show user sends obsolete owner=me query param and fails with 400
## Summary

`vastai show user` and `VastAI().show_user()` currently fail because they send `owner=me` to `GET /api/v0/users/current`. The API now rejects that extra query parameter.

The endpoint works when `owner` is omitted. This appears to be a recent API-side validation change, likely within the past 24 hours, because this same client path had been working before then.

## Environment

- Package: `vastai`
- Version tested: `1.0.7`
- Also present on current `master` as of `1850bbb00c73f61910d68cd0380a18f8af00aa61`
- Platform tested: Windows / Python 3.14

## Reproduction

```bash
uvx --from vastai vastai --version
# 1.0.7

uvx --from vastai vastai show user --raw --explain
```

The CLI prepares this request:

```text
GET https://console.vast.ai/api/v0/users/current?owner=me&api_key=...
```

It returns:

```json
{
  "error": true,
  "status_code": 400,
  "msg": "owner: Extra inputs are not permitted"
}
```

The SDK path fails the same way:

```python
from vastai import VastAI

client = VastAI()
client.show_user()
```

This raises:

```text
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url:
https://console.vast.ai/api/v0/users/current/?owner=me&api_key=...
```

## Expected behavior

`show user` should call `/api/v0/users/current` without `owner=me`.

A direct request without `owner` succeeds:

```text
GET https://console.vast.ai/api/v0/users/current?api_key=...
```

The response includes the expected user/account fields, including `balance`.

## Current source

Current `master` still sends `owner=me` here:

```python
# vastai/api/billing.py
def show_user(client):
    r = client.get("/users/current", query_args={"owner": "me"})
    r.raise_for_status()
    user_blob = r.json()
    user_blob.pop("api_key", None)
    return user_blob
```

The legacy CLI path also does this:

```python
# vast.py
req_url = apiurl(args, "/users/current", {"owner": "me"})
```

The OpenAPI file in this repo documents `GET /api/v0/users/current/` with bearer auth and no `owner` query parameter.

## Suggested fix

Remove `owner=me` from the `show_user` request:

```python
r = client.get("/users/current")
```

or, if the client requires an explicit dict:

```python
r = client.get("/users/current", query_args={})
```

The same change likely needs to be applied to the legacy `vast.py` implementation for compatibility.

stackanthony:
Resolves issue #382. Our new webserver validation on this endpoint is causing issues with `show user`. The webserver expects there to not be a body in the request, but show_user is adding `owner: me` as a query param. This is an unconsumed parameter by the endpoint, safe to remove.

### DIFF
--- a/vastai/api/billing.py
+++ b/vastai/api/billing.py
@@ -250,7 +250,7 @@ def show_user(client):
     Returns:
         dict: User data (with api_key removed).
     """
-    r = client.get("/users/current", query_args={"owner": "me"})
+    r = client.get("/users/current"})
     r.raise_for_status()
     user_blob = r.json()
     user_blob.pop("api_key", None)
--- a/vastai/api/billing.py
+++ b/vastai/api/billing.py
@@ -250,7 +250,7 @@ def show_user(client):
     Returns:
         dict: User data (with api_key removed).
     """
-    r = client.get("/users/current"})
+    r = client.get("/users/current")
     r.raise_for_status()
     user_blob = r.json()
     user_blob.pop("api_key", None)


vastzuby (approved):
lgtm

LucasArmandVast (approved):

PR #385, opened 2026-04-29
fix: show user issue in CLI

stackanthony:


AidaLadybug (approved):

vastzuby:
@stackanthony btw the `vast.py` is deprecated. we just left it there for backwards compatibliity

PR #388, opened 2026-05-04
fix: crypto pillow versions

stackanthony:


LucasArmandVast (approved):
Passes serverless smoke test

PR #389, opened 2026-05-05
add commands to start/stop deployments

vastzuby:
not gonna merge until next webserver release (May 11th) when the new start/stop endpoints go live


vastzuby:
already implemented

PR #392, opened 2026-05-06
fix(cli): preserve rentable=any in search offers body (#391)

### LINKED ISSUE #391: search offers silently strips `rentable` filter — can't query rented offers via CLI
# `search offers` silently strips `rentable` from request body — can't query rented offers via CLI

**vastai-cli version:** 1.0.9 (current PyPI release)
**Python:** 3.10
**OS:** Ubuntu 22.04

## What I expected

```
vastai search offers 'rented=true rentable=any gpu_name="RTX PRO 6000 WS"' --raw
```

…to return all currently-rented RTX PRO 6000 WS offers. The same JSON body sent
directly via `curl -X POST /api/v0/bundles/` returns 27 offers; via the CLI it
returns 0.

## What actually happens

`--explain` shows the CLI silently drops `rentable` from the request body and
auto-injects defaults:

```
$ vastai search offers 'rented=true rentable=any gpu_name="RTX PRO 6000 WS"' --explain --raw
request json:
{'verified': {'eq': True}, 'external': {'eq': False}, 'rented': {'eq': True},
 'gpu_name': {'eq': 'RTX PRO 6000 WS'}, 'order': [['score', 'desc']],
 'type': 'on-demand', 'allocated_storage': 5.0}
```

Note the absence of any `rentable` key. The same query via curl with
`{"rented":{"eq":true},"rentable":{"eq":"any"},"gpu_name":{"eq":"RTX PRO 6000 WS"}}`
returns the rented offers correctly.

`verified` and `external` defaults can be overridden by passing
`verified=any external=any` in the filter string, but adding `rentable=any`
or `rentable=true` to the filter has no effect — the key never reaches the
request body.

## Why it matters

For host operators doing competitive pricing intelligence, the rented-offer
view is essential — it shows what renters are *actually paying*, not just
the leftover unrented inventory. Without it, the CLI gives a misleadingly
empty picture of the market.

## Proposed fix

Add `rentable` to the list of CLI-exposed filters in
`vast/search.py` (or wherever `search offers` builds its request body),
alongside `rented`, `verified`, `external`, etc. Default to `rentable=true`
to preserve current behaviour; allow `rentable=any` and `rentable=false`
to surface rented and decommissioned offers respectively.

## Workaround

```
curl -X POST 'https://console.vast.ai/api/v0/bundles/' \
  -H "Authorization: Bearer $VAST_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"limit":500, "rented":{"eq":true}, "rentable":{"eq":"any"},
       "gpu_name":{"eq":"RTX PRO 6000 WS"}, "num_gpus":{"eq":1}}'
```

cparmar:
## Summary

`vastai search offers` silently strips `rentable=any` (and any other seeded-default field used with the `=any` wildcard), causing queries like `rented=true rentable=any` to return 0 even when matching rented offers exist. Fixes #391.

## Root cause

`vastai/cli/commands/offers.py:search__offers` seeds default filters before parsing the user query:

```python
query = {"verified": {"eq": True}, "external": {"eq": False},
         "rentable": {"eq": True}, "rented": {"eq": False}}
query = parse_query(args.query, query, ...)
```

`parse_query()` treats `field=any` / `field=*` as a wildcard and DELETES the key from the body (this is the documented behaviour — see `tests/api/test_query.py::test_wildcard_any_deletes_field`). Combined with the API server's own implicit `rentable=true` default, this means:

```
vastai search offers 'rented=true rentable=any gpu_name="RTX PRO 6000 WS"'
```

silently sends a body with `rentable` MISSING, the server re-applies `rentable=true`, and the empty intersection `(rented=true) AND (rentable=true)` returns 0 — while the same JSON sent via direct `curl POST /api/v0/bundles/` with `rentable={"eq": "any"}` set explicitly returns the rented offers.

## `--explain` trace from the broken case

```
$ vastai search offers 'rented=true rentable=any gpu_name="RTX PRO 6000 WS"' --explain --raw
request json:
{'verified': {'eq': True}, 'external': {'eq': False}, 'rented': {'eq': True},
 'gpu_name': {'eq': 'RTX PRO 6000 WS'}, 'order': [['score', 'desc']],
 'type': 'on-demand', 'allocated_storage': 5.0}
```

(`rentable` is absent — that is the bug.)

## Fix

Remember which keys the CLI seeded, and after `parse_query()` re-insert `{"eq": "any"}` for any seeded key the user wildcarded out. The server interprets `{"eq": "any"}` as "no filter on this field", overriding its implicit default. Scope is limited to the CLI handler — SDK callers already pass `no_default=True` after seeding so they take a different (compatible) path.

Patch is ~13 lines in `vastai/cli/commands/offers.py`. No refactor, no behaviour change for queries that don't use `=any` on a seeded field.

## Tests

Added two regression tests in `tests/cli/test_offers_commands.py`:

- `test_search_offers_rentable_any_preserved` — asserts that `rented=true rentable=any` lands in the request body as `{"rentable": {"eq": "any"}, "rented": {"eq": True}}`, NOT with `rentable` missing.
- `test_search_offers_seeded_defaults_preserved_when_unmentioned` — asserts that defaults the user never touches still ride through with their seeded values (regression guard).

## Test plan

- [x] Verify `parse_query` strips `=any` keys (existing test `test_wildcard_any_deletes_field`).
- [x] Manually run the patched function logic against the failing query and confirm `rentable={"eq": "any"}` ends up in the body.
- [ ] Maintainer: run `pytest tests/cli/test_offers_commands.py tests/api/test_query.py` in CI to confirm the two new tests pass and no existing tests regress.
- [ ] Maintainer: re-run the original failing CLI command against production and confirm it now returns the rented offers.

### DIFF
--- a/tests/cli/test_offers_commands.py
+++ b/tests/cli/test_offers_commands.py
@@ -35,6 +35,36 @@ class TestSearchOffers:
         captured = capsys.readouterr()
         assert "ID" in captured.out
 
+    def test_search_offers_rentable_any_preserved(self, parse_argv, patch_get_client, mock_response):
+        """Regression for #391: `rentable=any` must not be silently stripped.
+
+        Without the fix the field disappears from the body (parse_query
+        deletes `=any` keys), and the API server re-applies its implicit
+        `rentable=true` default — so `rented=true rentable=any` returns 0.
+        The CLI must instead send `{"rentable": {"eq": "any"}}` to override
+        the server-side default.
+        """
+        patch_get_client.post.return_value = mock_response(200, {"offers": []})
+        args = parse_argv(["search", "offers", "--raw", "rented=true rentable=any"])
+        args.func(args)
+        patch_get_client.post.assert_called_once()
+        json_data = patch_get_client.post.call_args[1]["json_data"]
+        assert json_data["rentable"] == {"eq": "any"}, (
+            f"rentable should be present as {{'eq': 'any'}} but body was: {json_data}"
+        )
+        assert json_data["rented"] == {"eq": True}
+
+    def test_search_offers_seeded_defaults_preserved_when_unmentioned(self, parse_argv, patch_get_client, mock_response):
+        """Defaults the user does NOT touch must remain at their seeded values."""
+        patch_get_client.post.return_value = mock_response(200, {"offers": []})
+        args = parse_argv(["search", "offers", "--raw", "num_gpus=1"])
+        args.func(args)
+        json_data = patch_get_client.post.call_args[1]["json_data"]
+        assert json_data["verified"] == {"eq": True}
+        assert json_data["external"] == {"eq": False}
+        assert json_data["rentable"] == {"eq": True}
+        assert json_data["rented"] == {"eq": False}
+
 
 class TestSearchTemplates:
     def test_search_templates(self, parse_argv, patch_get_client, mock_response, capsys):
--- a/vastai/cli/commands/offers.py
+++ b/vastai/cli/commands/offers.py
@@ -137,8 +137,21 @@ def search__offers(args):
         else:
             query = {"verified": {"eq": True}, "external": {"eq": False}, "rentable": {"eq": True}, "rented": {"eq": False}}
 
+        # Remember which keys we seeded so we can detect user removals via `=any`.
+        # parse_query() deletes a field when the user writes `field=any` (or `=*`).
+        # If we just let those keys disappear, the API server re-applies its own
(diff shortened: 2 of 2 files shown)


vastzuby:
thanks for reporting this bug @Chintan . it should be fixed if you update your package to v1.0.12 https://pypi.org/project/vastai/

though the CLI stripping rentable=any is intended and means "dont use this filter". the field stays in the body if rentable=true or false
your issue was a bug with the rented filter.

let me know if you run into any other issues 👍

(by the way rented=true means include offers that you're currently renting)

PR #399, opened 2026-05-13
self-test: route to cu118/cu128/cu130 images with compute_cap aware fallback

robballantyne:
## Summary

Update `self_test__machine`'s docker image selection to match the new three-image lineup (`vastai/test:self-test-cuda-11.8|12.8|13.0`) and add a compute-capability rule so older GPUs land on a wheel that actually has kernels for them. Also fixes the test-instance cleanup so it actually runs on Ctrl+C.

## Image selection

`docker_tag_map` is trimmed to `cuda-11.8 / cuda-12.8 / cuda-13.0` (drops the now-unpublished `cu121 / cu124 / cu126` entries). The existing decrement-by-0.1 fallback still resolves intermediate CUDA versions to the nearest lower tag — e.g. a host on CUDA 12.4 steps down to `cuda-11.8`, since the host driver can't load `cuda-12.8` wheels.

Two `compute_cap` branches added on top:

1. **`compute_cap < 700` → force `cuda-11.8`.** `cuda-12.8` (torch 2.10) and `cuda-13.0` (torch 2.11) wheels don't compile `sm_50`/`sm_60` kernels, so Maxwell and Pascal must use the legacy image regardless of CUDA version.
2. **`compute_cap < 750` (Volta) + `cuda_max_good > 12.8` → clamp `cuda_version` to `12.8`.** `cuda-12.8` (pinned to torch 2.10) still ships `sm_70` kernels, but `cuda-13.0` wheels never did. Without the clamp, a V100 with a CUDA 13 driver installed would resolve to `cuda-13.0` and fail with "no kernel image available." Cap and let the version map pick `cuda-12.8`.

Sample decisions (now logged with the reason):

| Host | Image | Reason |
|---|---|---|
| V100 + CUDA 13 driver | `cuda-12.8` | `compute_cap=700 (Volta) + cuda_max_good=13.0 → clamped to 12.8 → cuda-12.8` |
| V100 + CUDA 12.8 | `cuda-12.8` | `compute_cap=700, cuda_max_good=12.8 → exact match → cuda-12.8` |
| P100 (Pascal) | `cuda-11.8` | `compute_cap=600 below sm_70 → forced cuda-11.8` |
| H100 + CUDA 12.9 | `cuda-12.8` | `compute_cap=900, cuda_max_good=12.9 → stepped down to 12.8 → cuda-12.8` |
| B200 + CUDA 13 | `cuda-13.0` | `compute_cap=1000, cuda_max_good=13.0 → exact match → cuda-13.0` |

`cuda_map_to_image` now returns `(image, reason)` and the `Starting test with …` progress line prints both, so operator logs say *why* a particular image was chosen.

## Ctrl+C cleanup

`KeyboardInterrupt` inherits from `BaseException`, not `Exception`, so the outer `except Exception` was letting it through and the cleanup block's `except Exception: pass` silently ate any resulting destroy failure. Hitting Ctrl+C during the wait/poll phase could leak the test instance.

- Added an explicit `except KeyboardInterrupt` branch above `except Exception` that sets the result reason and prints a cleanup message. The `finally` still runs to do the destroy.
- Dropped the `except Exception: pass` in the cleanup. If the destroy call fails for any reason — auth, network, transient API — a `WARNING:` line now names the instance id and prints the `vastai destroy instance <id>` recovery command.
- The cleanup also catches its own `KeyboardInterrupt` (a second Ctrl+C during the destroy API call) and prints the same manual-cleanup hint before re-raising.

Identical changes applied to the monolithic `vast.py` (`self_test__machine`) and the modular `vastai/cli/commands/machines.py`.

## Test plan

- [ ] `vastai self-test machine <id>` on a Turing+ host with CUDA 12.8 driver — expect `cuda-12.8`, exact match line.
- [ ] `vastai self-test machine <id>` on a V100 with CUDA 13 driver — expect `cuda-12.8`, clamped line.
- [ ] `vastai self-test machine <id>` on a Pascal host — expect `cuda-11.8`, forced line.
- [ ] During an in-progress self-test, hit Ctrl+C. Confirm the test instance is destroyed and no leaked instance shows up in `vastai show instances`.
- [ ] Run with the destroy intentionally broken (e.g. revoked API key) and confirm the new WARNING surfaces the instance id and the manual cleanup command.

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -8417,15 +8417,15 @@ def self_test__machine(args):
             """
             Map a CUDA version (and optional compute_cap) to (image, reason).
 
-            If compute_cap is below 700 (sm_70 / Volta), the cu118 image is
-            forced regardless of CUDA version. cu128 (torch 2.10) still ships
-            sm_70 kernels so Volta hosts land on cu128 via the version map;
-            cu130 (torch 2.11) and cu128 do not ship sm_50/sm_60 kernels, so
-            Maxwell and Pascal must use cu118.
+            If compute_cap is below 700 (sm_70 / Volta), the cuda-11.8 image
+            is forced regardless of CUDA version. cuda-12.8 (torch 2.10) still
+            ships sm_70 kernels so Volta hosts land on cuda-12.8 via the
+            version map; cuda-13.0 (torch 2.11) and cuda-12.8 do not ship
+            sm_50/sm_60 kernels, so Maxwell and Pascal must use cuda-11.8.
 
             Volta hosts whose operator has installed a CUDA 13 driver get the
             cuda_version clamped down to 12.8 before the version map runs,
-            since cu130 wheels never built sm_70.
+            since cuda-13.0 wheels never built sm_70.
 
             The returned reason is a short human-readable string so callers
             can log why a particular image was chosen.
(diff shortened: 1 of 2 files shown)


robballantyne:
Existing image tags will be re-pointed at new images only after merge

hanran-y (approved):

PR #400, opened 2026-05-13
tell user which api key used on 401 error, group --help commands

vastzuby:
two changes 

1. group commands in --help 

2. if user enables TFA, their api key just stops working. so now wetell the user why it failed and how to 2fa login through the cli

3. api key issue 
## issue 
users hit 401 "invalid user key" even after calling `vastai set api-key`. 

## why
the cli is using their shell key. we give no clue to the user which key is actually sent. 

Key order is
1. --api-key flag
2. $VAST_API_KEY
3. file written by `set api-key`

## fix 
`vastai set api-key` now warns if the $VAST_API_KEY is set in the env. since the saved key wont be used until thats unset 
we also print the keys to the user (only last 4 digits) so they can compare their keys and remove the stale one

### DIFF
--- a/vastai/cli/commands/auth.py
+++ b/vastai/cli/commands/auth.py
@@ -8,7 +8,7 @@ from datetime import datetime, timezone
 from vastai.cli.parser import argument
 from vastai.cli.display import deindent, display_table
 from vastai.api import auth as auth_api
-from vastai.cli.util import SUCCESS, WARN, FAIL
+from vastai.cli.util import SUCCESS, WARN, FAIL, fmt_key_suffix
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


sammy-vastai:
Shouldn't set api key update `$VAST_API_KEY`?  Basically if we call that command, we should set set the api key in the .config.  When reading, always read from the .config before the .env.  The env is more legacy and config is the newer standard.

LucasArmandVast (approved):

LucasArmandVast:
> Shouldn't set api key update `$VAST_API_KEY`? Basically if we call that command, we should set set the api key in the .config. When reading, always read from the .config before the .env. The env is more legacy and config is the newer standard.

The reasoning was that the .config was fixed, but the .env is temporary for this shell session, so the .env behaves more like an override and the .config is better for a permanent setting. Consider the case where we want to use a different API key temporarily; we can do:

```VAST_API_KEY="temp_key" vastai search offers ...```.

We could directly update the env when we call set api key, but then this would only last for duration of the current shell session. We shouldn't mess with the users .bashrc/.zshrc or other shell config for them.

PR #401, opened 2026-05-16
feat(CLN-3107): New product direction price increase accept with Observability and visibility

abiola-vastai:
This replaces the previous implementation of price-increase-extension flow with a simpler pending-row architecture. host edits upsert one pending_price_increases row per affected contract; the client lists pending rows and accepts/rejects each by pending_price_increase_id. Drops the abandoned successor-row + auth-challenge + email-token direction along with its 2102+ line price_increase_extensions.py.


Marcoh572:
Really like this rework — the per-row model reads much cleaner than the old batch/`--host` shape, and the docstrings pinning the exact backend contract (body shape, 404/409, the cutover-after-`end_date` note) will genuinely help the next person. A few small things, none blocking:

- exit codes: the 0/1/2 split (2 = stale) is nice for scripting — is a differentiated exit-2 a convention we want to set CLI-wide? Most other commands just exit 0 or let errors bubble, so flagging in case we'd rather keep it uniform. Totally fine if it's intentional.
- duplicate ids: `accept price-increase 123 123` accepts once then reports the second as stale (exit 2), and the confirm prompt still fires even when none of the ids have a pending row. Could dedup the ids and skip the prompt when nothing matches — minor though.
- a few micro things: `set(instance_ids)` gets rebuilt each row inside the `_prompt_or_exit` comprehension (could hoist it out); `accepted_ids` in `_run_per_row` actually holds contract_ids on the reject path too, so the name's a touch off there; and conftest's `COMMAND_MODULES` lists `benchmarks` twice.

Left a few inline notes too. Nice work overall!

vastai/sdk.py
```diff
@@ -106,24 +106,46 @@ def change_bid(self, id: int, price: Optional[float] = None) -> dict:
...
+        from vastai.api import price_increase
+        envelope = price_increase.list_pending(self.client)
+        rows = envelope.get("pending_price_increases", []) or []
+        match = next((r for r in rows if r.get("contract_id") == int(instance_id)), None)
+        if match is None:
+            raise LookupError(
+                f"no pending price increase for instance {instance_id}")
+        return price_increase.accept(self.client, match["pending_price_increase_id"])
+
+    def reject_price_increase(self, instance_id: int) -> dict:
```
Marcoh572, line 130: Small thing — this "list pending → match on `contract_id` → grab `pending_price_increase_id`" dance shows up here, in `accept_price_increase` just above, and in the `api/instances.accept_price_increase` shim. Might be worth a single helper (`_resolve_instance_to_pending(client, instance_id)` that returns the row or raises `LookupError`) so a change to the match key is one edit instead of three. Not urgent.

tests/cli/test_price_increase_commands.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def test_storage_with_null_new_shows_dash(self, parse_argv, patch_get_client, mock_response, capsys):
+        patch_get_client.get.return_value = mock_response(200, _envelope([PENDING_ROW]))
+        args = parse_argv(["show", "pending-price-increases"])
+        args.func(args)
+        captured = capsys.readouterr()
+        # `new_disk_ram_costpersec` is None on the fixture, so the storage cell
+        # collapses to "-" (rendered as "-" in display_table; "_" if space).
+        # Either "_" (replaced spaces) or "-" should be present.
+        assert " - " in captured.out.replace("_", " ") or "-" in captured.out
```
Marcoh572: I could be wrong, but I think this assertion passes regardless — `"-" in captured.out` is basically always true since the usage/header text contains "price-increase" (plenty of dashes). So it wouldn't catch a regression where the storage cell stopped collapsing to "-". Maybe assert on the rendered Storage column specifically (that the cell for this row is "-")? The intent is great, just want it to have teeth.

vastai/cli/commands/price_increase.py
```diff
@@ -0,0 +1,337 @@
...
+        display_table(
+            [_prepare_row(r) for r in matching], pending_price_increase_fields,
+        )
+    answer = input(
+        f"{present_imperative} these price increases? [y/N]: "
+    ).strip().lower()
+    return answer == "y"
+
+
+def _run_per_row(args, *, route_verb, api_call, present_imperative, past_tense, cutover_note):
```
Marcoh572, line 232: Should `--raw` do something here? `show pending-price-increases` honours it (returns the envelope so `main` prints JSON), but on accept/reject `_run_per_row` returns `None` and prints the human lines — so `accept price-increase 123 --raw` still emits prose to stdout, and a script piping it to `jq` would choke. The old command had `if args.raw: return rj`. Maybe collect the per-row results into a list and return it when `args.raw` so the output stays machine-readable? Could be intentional given the exit codes already carry the signal — your call.

vastai/api/price_increase.py
```diff
@@ -0,0 +1,89 @@
...
+    r = client.get("/instances/pending-price-increases/")
+    r.raise_for_status()
+    return r.json()
+
+
+def accept(client: VastClient, pending_id: int) -> dict:
+    """Accept one pending price-increase row.
+
+    The body must be exactly ``{"pending_price_increase_id": int}``;
+    backend ``extra='forbid'`` returns 422 for any extra key. Returns
```
Marcoh572: Tiny doc nit: I think this is a 400, not 422 — the param-validation tween (`web/utils/security_utils.py:537,548`) returns `HTTPBadRequest` with `{"success": false, "error": "bad_request"}` on a pydantic `ValidationError`. Doesn't change anything in practice since we never send extra keys, but the docstrings here are so precise about the contract that it's worth matching. Same "422" shows up in the module docstring and the `tests/api/test_price_increase.py` header.

Marcoh572:
Nice round — the dedup + skip-prompt-on-no-matches logic, the `resolve_instance_to_pending` extraction, the `_rendered_table_cell` helper with the ANSI strip, and the 400-not-422 catches everywhere they needed to land all look good. Left one inline on the new `--raw` machinery.

vastai/cli/commands/price_increase.py
```diff
@@ -0,0 +1,407 @@
...
+            "contract_id": result.get("contract_id"),
+            "status": past_tense.lower(),
+            "response": result,
+        })
+        acted_ids.append(result.get("contract_id"))
+
+    total = len(instance_ids)
+    n_ok = len(acted_ids)
+    if args.raw:
+        exit_code = 1 if failed else 2 if stale else 0
```
Marcoh572, line 307: One mechanism thing on the new `--raw` path — the returned dict carries `exit_code: 1/2/0`, but `main.py:140-145` always `sys.exit(0)` after printing JSON for any `args.raw` command, so `accept price-increase 123 --raw` on a 500 will print `{"exit_code": 1, …}` to stdout but the process still exits 0. A shell pipeline or CI runner only seeing `$?` would think it succeeded — which is exactly the scripting case the `--raw` rework was meant to enable.

Could do `print(json.dumps(payload, indent=1, sort_keys=True)); sys.exit(exit_code)` here in the raw branch (and return `None`) so the process exit matches the dict's claim — keeps the fix contained to this command instead of teaching `main` to special-case dict returns.

Tiny side-note: the dict also has both `"exit_code"` and `"_exit_code"` (same value) — looks like a half-finished hook; whichever way you go on the above, dropping the underscored one would clean that up. Also worth a test covering the non-happy raw path (e.g. PUT 500 → exit code 1) — the existing `test_raw_returns_machine_readable_summary` only walks the happy path.

abiola-vastai: Check now

Marcoh572 (approved):
R3: --raw exit-code mechanism is correct now, payload shape clean, failure-path test exercises the real process exit via pytest.raises(SystemExit). LGTM.

PR #405, opened 2026-05-27
Handle missing instances during self-test cleanup

jjziets:
## Summary

Fixes the cleanup warning found during CON-1509 self-test dogfooding.

The live stock self-test passed and destroyed its test instance, but the CLI then checked the already-deleted instance again. The API returned `{"instances": null}`, `show_instance()` crashed with `TypeError: 'NoneType' object is not subscriptable`, and self-test printed a false manual-destroy warning even though `show instances` showed zero active instances.

Changes:

- Treat `{"instances": null}` as a missing/deleted instance in `show_instance()`.
- Make `vastai show instance <id>` return clean output for deleted instances instead of traceback.
- Make self-test final cleanup treat a missing instance as already gone, without duplicate destroy warnings.
- Keep SDK `ssh_url` / `scp_url` safe when `show_instance()` returns `None`.
- Add focused CLI and SDK tests.

## Verification

- `uv run pytest tests/cli/test_instances_commands.py tests/cli/test_machines_commands.py tests/sdk/test_sdk.py -q`
- Read-only repro now returns clean raw JSON instead of traceback:
  `vastai --raw show instance 38056189` -> `{ "instances": null }`

## Notes

This PR only addresses the cleanup/deleted-instance bug. It does not change self-test image selection or the private `vast-ai/self-test` image sources.

### DIFF
--- a/tests/cli/test_instances_commands.py
+++ b/tests/cli/test_instances_commands.py
@@ -44,6 +44,24 @@ class TestShowInstance:
         call_args = patch_get_client.get.call_args
         assert "123" in call_args[0][0]
 
+    def test_show_instance_raw_deleted_instance(self, parse_argv, patch_get_client, mock_response):
+        patch_get_client.get.return_value = mock_response(200, {"instances": None})
+        args = parse_argv(["show", "instance", "123", "--raw"])
+
+        result = args.func(args)
+
+        assert result == {"instances": None}
+
+    def test_show_instance_display_deleted_instance(self, parse_argv, patch_get_client, mock_response, capsys):
+        patch_get_client.get.return_value = mock_response(200, {"instances": None})
+        args = parse_argv(["show", "instance", "123"])
+
+        result = args.func(args)
+
+        assert result == 1
+        captured = capsys.readouterr()
+        assert "Instance 123 not found or no longer exists." in captured.out
+
 
 class TestDestroyInstance:
     def test_destroy_instance(self, parse_argv, patch_get_client, mock_response, capsys):
--- a/tests/cli/test_instances_commands.py
+++ b/tests/cli/test_instances_commands.py
@@ -20,6 +20,14 @@ class TestShowInstances:
         assert "/instances" in call_args[0][0]
         assert isinstance(result, list)
 
+    def test_show_instances_raw_null_instances(self, parse_argv, patch_get_client, mock_response):
+        patch_get_client.get.return_value = mock_response(200, {"instances": None})
+        args = parse_argv(["show", "instances", "--raw"])
+
+        result = args.func(args)
+
+        assert result == []
+
     def test_show_instances_display(self, parse_argv, patch_get_client, mock_response, capsys):
         patch_get_client.get.return_value = mock_response(200, {
             "instances": [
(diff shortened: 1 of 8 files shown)


jjziets:
Fresh validation from the pushed GitHub branch:

- Fresh clone: `jjziets/vast-python`, branch `fix/self-test-cleanup-missing-instance`, commit `97b062a`.
- Focused tests passed: `python -m pytest tests/cli/test_instances_commands.py tests/cli/test_machines_commands.py tests/sdk/test_sdk.py -q` -> `84 passed`.
- Deleted-instance repro now returns clean raw JSON with no traceback: `vastai --raw show instance 38056189` -> `{ "instances": null }`.
- Live dogfood from the fresh clone: `vastai self-test machine 46368` created instance `38061885`, passed, destroyed the instance on attempt 1, and did not print the previous false cleanup warning.
- Final audit after the live test: `vastai --raw show instances` returned zero active instances.

robballantyne (changes requested):
Tested locally on the branch — thanks for the fix.
  
  Verified working:
  - vastai show instance <deleted-id> --raw → clean {"instances": null}, no traceback
  - vastai show instance <deleted-id> → friendly Instance X not found or no longer exists. with exit code 1 (parser dispatch does honor the
  returned int)
  
  Worth addressing: legacy vast.py is not patched

  The same crash still exists at vast.py:5702:

  row = r.json()["instances"]
  row['duration'] = time.time() - row['start_date']   # TypeError if row is None

  Most users install the published vastai package, so the blast radius is small. But the project still documents downloading the monolithic
  vast.py as an alternative install method, so users on that path hit the unfixed crash.

  Recent history suggests we patch both trees for bug fixes: 4 of the last 37 commits touched both vast.py and vastai/, and they were all bug
   fixes / cross-cutting changes. The closest precedent is #399 (self-test image routing) from the same area, which updated both. Feature
  work is vastai/-only, but bugs generally aren't.

  Please also patch vast.py:5687 (show__instance) and the self-test cleanup block in vast.py to mirror the
  vastai/cli/commands/machines.py:1038 change.

  Smaller nits (non-blocking):

  - vastai/sdk.py:69 — def show_instance(self, id: int) -> dict now returns Optional[dict]; please align the annotation with
  vastai/api/instances.py.
  - vastai/cli/commands/instances.py:92 — the "not found" message goes to stdout; consider print(..., file=sys.stderr) so scripts piping
  stdout aren't polluted.
  - vastai/api/instances.py:31 (show_instances, plural) has the same r.json()["instances"] shape and would crash on null. Probably out of
  scope, but a one-line guard would keep the two functions symmetric.
  - The new tests/cli/test_machines_commands.py::TestSelfTestMachineCleanup test mocks the entire self-test pipeline (offers, create, three
  sequential show_instance returns, destroy, requests, sleep). It'll catch the current bug but is brittle to any refactor of the polling
  loop. A unit test exercising just the finally cleanup block would be more durable — happy to leave as-is if you'd prefer.

  Once vast.py is patched I think this is good to merge.

jjziets:
Thanks Rob, addressed.

Changes pushed in `3bf674f`:
- Patched legacy `vast.py` `show__instance` to handle `{"instances": null}` without traceback.
- Mirrored the self-test cleanup behavior in legacy `vast.py` so already-gone instances do not make cleanup crash/warn incorrectly.
- Updated `vastai/sdk.py` return annotation to `Optional[dict]`.
- Moved the modular CLI not-found message to stderr.
- Added a null guard for plural `show_instances`.
- Updated tests for stderr behavior and null plural responses.

Verified with:
```
./.venv/bin/python -m pytest tests/cli/test_instances_commands.py tests/cli/test_machines_commands.py
# 30 passed
```

robballantyne (approved):

PR #406, opened 2026-05-27
Warn when self-test ignores requirements

jjziets:
## Summary

Addresses the `--ignore-requirements` self-test messaging issue.

When this flag is used, the CLI now warns that requirement checks are skipped as a pass/fail gate and that passing the self-test does not qualify the machine for verification.

Changes:

- Adds an explicit warning whenever `--ignore-requirements` is set.
- Prints the warning before the live self-test starts.
- Prints the warning again in the final console status output.
- Includes the warning in raw JSON output as `warning`.
- Adds focused tests for successful console output and raw summary output.

## Verification

- `uv run pytest tests/cli/test_machines_commands.py tests/cli/test_util.py -q` -> `58 passed`

## Notes

This PR is intentionally separate from #405. PR #405 fixes the deleted-instance cleanup warning found during dogfood; this PR only handles `--ignore-requirements` messaging.

PR #407, opened 2026-05-27
CON-1512: Improve self-test ignore-requirements messaging

jjziets:
## Summary

Addresses CON-1512: `Self-Test: Improve self-test messaging when run with --ignore-requirements`.

When this flag is used, the CLI now warns that requirement checks are skipped as a pass/fail gate and that passing the self-test does not qualify the machine for verification.

Changes:

- Adds an explicit warning whenever `--ignore-requirements` is set.
- Prints the warning before the live self-test starts.
- Prints the warning again in the final console status output.
- Includes the warning in raw JSON output as `warning`.
- Adds focused tests for successful console output and raw summary output.

## Verification

- `uv run pytest tests/cli/test_machines_commands.py tests/cli/test_util.py -q` -> `58 passed`
- Fresh-clone validation from `jjziets/vast-python:fix/self-test-ignore-requirements-warning` before branch rename also passed: `58 passed`.
- Actual raw/no-rent CLI check from the fresh clone returned a `warning` field:
  ```json
  {
    "success": false,
    "reason": "No valid offers found.",
    "warning": "WARNING: --ignore-requirements is set. Requirement checks are skipped as a pass/fail gate, and passing this self-test does not qualify this machine for verification."
  }
  ```

## Notes

This supersedes closed PR #406, which GitHub closed when the fork branch was renamed to carry the CON-1512 branch name.

This PR is intentionally separate from #405. PR #405 fixes the deleted-instance cleanup warning found during dogfood; this PR only handles `--ignore-requirements` messaging.

### DIFF
--- a/tests/cli/test_machines_commands.py
+++ b/tests/cli/test_machines_commands.py
@@ -1,5 +1,6 @@
 """Integration tests for machine CLI commands with mocked HTTP."""
 
+import json
 from types import SimpleNamespace
 from unittest.mock import Mock
 
@@ -145,3 +146,66 @@ class TestListMachineEpilogDoesNotPromiseEmail:
 
     def test_list_machines_epilog_no_email(self, cli_parser):
         assert "email" not in self._epilog_lower(cli_parser, "list machines")
+
+
+class TestSelfTestMachineIgnoreRequirements:
+    def test_ignore_requirements_warns_on_success(self, parse_argv, monkeypatch, capsys):
+        from vastai.cli.commands import machines
+
+        offer = {
+            "id": 202,
+            "dlperf": 1.0,
+            "cuda_max_good": 13.0,
+            "compute_cap": 1200,
+            "reliability": 0.99,
+            "direct_port_count": 10,
+            "pcie_bw": 4.0,
+            "gpu_total_ram": 32 * 1024,
+            "inet_down": 200.0,
+            "inet_up": 200.0,
+            "gpu_ram": 32,
+            "cpu_ram": 64 * 1024,
+            "cpu_cores": 8,
+            "num_gpus": 1,
+        }
+        instance = {
+            "intended_status": "running",
+            "actual_status": "running",
+            "public_ipaddr": "203.0.113.10",
+            "ports": {"5000/tcp": [{"HostPort": "5000"}]},
+        }
+
+        monkeypatch.setattr(machines.offers_api, "search_offers", Mock(return_value=[offer]))
+        monkeypatch.setattr(machines.instances_api, "create_instance", Mock(return_value={"new_contract": 303}))
+        monkeypatch.setattr(machines.instances_api, "show_instance", Mock(return_value=instance))
+        monkeypatch.setattr(machines.instances_api, "destroy_instance", Mock(return_value={"success": True}))
+        monkeypatch.setattr(machines.requests, "get", lambda *_, **__: SimpleNamespace(status_code=200, text="DONE"))
+        monkeypatch.setattr(machines.time, "sleep", lambda *_: None)
+
+        args = parse_argv(["self-test", "machine", "123", "--ignore-requirements"])
+        with pytest.raises(SystemExit) as exc_info:
+            args.func(args)
+
+        assert exc_info.value.code == 0
+        out = capsys.readouterr().out
+        assert "WARNING: --ignore-requirements is set." in out
+        assert "Requirement checks are skipped as a pass/fail gate" in out
+        assert "does not qualify this machine for verification" in out
+        assert out.count("does not qualify this machine for verification") >= 2
+        assert "Test passed." in out
+
+    def test_ignore_requirements_warning_in_raw_summary(self, parse_argv, monkeypatch, capsys):
+        from vastai.cli.commands import machines
+
+        monkeypatch.setattr(machines.offers_api, "search_offers", Mock(return_value=[]))
+
+        args = parse_argv(["--raw", "self-test", "machine", "0", "--ignore-requirements"])
+        with pytest.raises(SystemExit) as exc_info:
+            args.func(args)
+
+        assert exc_info.value.code == 0
+        raw = json.loads(capsys.readouterr().out)
+        assert raw["success"] is False
+        assert "warning" in raw
+        assert "Requirement checks are skipped as a pass/fail gate" in raw["warning"]
+        assert "does not qualify this machine for verification" in raw["warning"]
--- a/vast.py
+++ b/vast.py
@@ -8374,10 +8374,17 @@ def self_test__machine(args):
     """
     instance_id = None  # Store instance ID for cleanup if needed
     result = {"success": False, "reason": ""}
+    ignore_requirements_warning = (
+        "WARNING: --ignore-requirements is set. Requirement checks are skipped as a "
+        "pass/fail gate, and passing this self-test does not qualify this machine for verification."
+    )
     
     # Ensure debugging attribute exists in args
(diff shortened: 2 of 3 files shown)


jjziets:
Additional validation after renaming the branch to `CON-1512-self-test-ignore-requirements-messaging`:

- Fresh clone: `jjziets/vast-python`, branch `CON-1512-self-test-ignore-requirements-messaging`, commit `99036bf`.
- Focused tests passed: `python -m pytest tests/cli/test_machines_commands.py tests/cli/test_util.py -q` -> `58 passed`.
- Actual raw/no-rent CLI check from the fresh clone:
  `vastai --raw self-test machine 0 --ignore-requirements | jq .`
  returned:
  ```json
  {
    "success": false,
    "reason": "No valid offers found.",
    "warning": "WARNING: --ignore-requirements is set. Requirement checks are skipped as a pass/fail gate, and passing this self-test does not qualify this machine for verification."
  }
  ```

jjziets:
Paid live validation from the CON-1512 branch (`CON-1512-self-test-ignore-requirements-messaging`, commit `99036bf`):

1. Known-good machine with `--ignore-requirements`
   - Command: `vastai self-test machine 46368 --ignore-requirements`
   - Result: self-test passed.
   - The new warning printed before the rented test started.
   - The new warning printed again in final output before `Test completed successfully.`
   - Instance `38064522` was destroyed; final account audit showed zero active instances.
   - Note: this branch is intentionally separate from #405, so the unrelated deleted-instance cleanup warning from master still appeared after destroy. That is covered by #405 and not part of CON-1512.

2. Requirements-failing low-upload machine
   - Machine: `39731` (Tesla P4, cheap, upload below requirement).
   - Without `--ignore-requirements`, the CLI stopped before renting: `Upload speed < 100 Mb/s`.
   - With `--ignore-requirements`, the CLI listed the unmet requirement, printed the new warning, and continued to rent instance `38065310`.
   - The instance stayed in `loading`, so I interrupted to avoid unnecessary spend; cleanup destroyed it.

3. Requirements-failing low-upload + low-RAM machine
   - Machine: `56491` (GTX 1660 S, upload below requirement, system RAM below total VRAM).
   - Without `--ignore-requirements`, the CLI stopped before renting and listed both unmet requirements.
   - With `--ignore-requirements`, the CLI listed both unmet requirements, printed the new warning, and continued to rent instance `38065994`.
   - The instance stayed in `loading`, so I interrupted to avoid unnecessary spend; cleanup destroyed it.

4. Requirements-failing no-direct-ports + low-bandwidth machine
   - Machine: `54894` (V100, direct_port_count=0, download/upload below requirement).
   - Without `--ignore-requirements`, the CLI stopped before renting and listed direct-port and bandwidth failures.
   - With `--ignore-requirements`, the CLI listed all unmet requirements, printed the new warning, and continued to rent instance `38066393`.
   - The instance stayed in `loading`, so I interrupted to avoid unnecessary spend; cleanup destroyed it.

Final audit after all live tests:

- `vastai --raw show instances` returned zero active instances.

This validates the key CON-1512 behavior: `--ignore-requirements` no longer quietly looks like normal qualification. It visibly says the requirements are skipped as a pass/fail gate and that passing does not qualify the machine for verification, including in final output and raw JSON.

robballantyne (changes requested):
When vastai self-test machine ... --ignore-requirements is used, the user now gets an explicit warning that requirement checks are no
  longer a pass/fail gate and that passing the test does not qualify the machine for verification. The warning lands in three places:

  - After check_requirements() returns (console only, via progress_print)
  - Before the final "Test completed/failed" line (console only)
  - Inside the raw JSON output as result["warning"]

  Net +78/-0 across 2 files (one source, one test). Tests cover both console and raw paths.

  What's good

  - User-facing improvement is real — the previous behavior silently downgraded the gate without informing the user, which is exactly the
  kind of footgun this warning addresses.
  - Single source of truth for the warning string — defined once at machines.py:553 and reused, so there's no drift between console and raw
  output.
  - Raw output gets the warning as a structured field (result["warning"]), not just a string baked into the human-readable message. Good for
  downstream tooling.
  - Tests cover both happy-path success with warning printed and raw-summary with warning embedded.

  Issues / suggestions

  - vast.py is not patched (vast.py:8363, vast.py:8414). The legacy monolith has its own self_test__machine with identical
  --ignore-requirements handling — same exact wording for "Continuing despite unmet requirements because --ignore-requirements is set." Users
   who installed via the documented vast.py-download path won't see the new warning. Recent project history shows bug/UX fixes in shared
  command surface get mirrored in both trees (e.g. #399 self-test image routing). Same observation as #405.
  - Inconsistent attribute access (machines.py:556 vs machines.py:670):
  if getattr(args, "ignore_requirements", False):   # line 556 — defensive
      result["warning"] = ignore_requirements_warning
  ...
  if args.ignore_requirements:                       # line 670 — direct
      progress_print(ignore_requirements_warning)
  - Argparse always provides the attribute (default False), so the getattr is harmless but stylistically inconsistent with the rest of the
  function. Pick one.
  - Warning fires even when requirements were genuinely met. The new if args.ignore_requirements: progress_print(warning) at machines.py:670
  is outside the if not meets_requirements branch, so a machine that passes requirements still gets the warning. Arguably correct (the user
  did pass the flag, so they should know what it means), but worth confirming this is the intended UX. Alternative: only warn when the flag
  actually bypassed something.
  - Warning printed twice in console output. Once at machines.py:671 (after check_requirements) and again at machines.py:1071 (before "Test
  completed successfully."). The test asserts >= 2 on the warning string count, so this is deliberate. For long self-test runs, the warning
  at the start gets scrolled off, so the repeat at the end is reasonable. For shorter runs it may feel noisy. Up to you — I'd keep both.
  - Test brittleness (tests/cli/test_machines_commands.py::TestSelfTestMachineIgnoreRequirements). test_ignore_requirements_warns_on_success
  mocks search_offers, create_instance, show_instance, destroy_instance, requests.get, and time.sleep — same heavy-mock pattern as the new
  test in #405. It catches the current behavior but is fragile to refactors of the polling loop. A more focused unit test that exercises only
   the warning-emission paths (not the full pipeline) would be more durable. Non-blocking.
  - assert "Test passed." in out relies on the substring printed by run_machinetester at machines.py:950, not the final summary line. Works
  today; brittle to wording changes. Worth a comment in the test noting where it originates.

  Risk

  Low. Pure additive UX change — no logic on the test pipeline itself is altered, only new prints and a new dict key. Worst case is
  duplicate/noisy warnings, not broken behavior.

  Verdict

  LGTM on the vastai/ change. The one thing I'd ask before merge is to also patch vast.py:8363/vast.py:8414 so the warning ships to
  monolith-install users too, mirroring the precedent set by #399. The style/test nits are non-blocking.

jjziets:
Thanks Rob, addressed.

Changes pushed in `678c31c`:
- Mirrored the `--ignore-requirements` warning into legacy `vast.py`.
- Added the warning to raw JSON output via `result["warning"]`.
- Printed the warning before the live self-test starts and again before the final success/failure summary.
- Fixed the legacy indentation so the “Continuing despite unmet requirements...” line prints once, not once per unmet reason.
- Cleaned the modular code to use `args.ignore_requirements` directly.

Verified with:
```
./.venv/bin/python -m pytest tests/cli/test_machines_commands.py tests/cli/test_util.py -q
# 58 passed
```

Dogfooded the no-rent path for both modular and legacy CLIs. The local saved API key returned 401, so this was not a paid/live rental validation, but it did verify the warning appears in raw and normal output for both trees.

robballantyne (approved):

Issue #1, opened 2019-10-01
Make into python package


ghandic:
Hey guys, just looking at using your script to run some goodies on GPUs. Rather than copying the script and making it executable and having it sat around somewhere without version control etc. It might be a good idea to make this into a pip package and store it on PyPI for people. Just my 2 cents - happy to submit a PR

ghandic:
Made a few changes, and made it into a package that can be put onto PyPi (obviously need to change author but add me as a contributor) https://github.com/ghandic/vast-python/

PR #408, opened 2026-05-27
CON-1510: Improve self-test diagnostics and runtime errors

jjziets:
## Summary

This draft implements the CLI-side CON-1510 self-test diagnostics work:

- add structured preflight diagnostics with actual values, thresholds, purpose, and remediation;
- render preflight failures once instead of duplicating the block;
- replace stale/literal remediation commands with current `search offers` syntax;
- preserve `--raw` compatibility while adding `diagnostics_version`, `checks`, `failure`, `failure_code`, `stage`, `error`, and offer metadata;
- reuse one selected offer for preflight and rental;
- add `--test-image` and `VAST_SELF_TEST_IMAGE` so rebuilt self-test images can be dogfooded before official tags are changed;
- add runtime failure codes and a legacy progress text parser for current images;
- classify create/start/progress/daemon failures with host-visible underlying errors and remediation;
- redact API keys from host-visible exception output;
- avoid the unnecessary Jupyter runtime path for self-test instances by launching with `ssh_direc ssh_proxy`.

This branch currently includes the deleted/missing-instance cleanup handling from PR #405 because CON-1510 cleanup diagnostics rely on the same behavior. Happy to rebase once #405 lands.

## Validation

Focused tests:

```text
uv run --with pytest --with pycryptodome python -m pytest tests/cli/test_machines_commands.py tests/cli/test_runtime_diagnostics.py tests/cli/test_instances_commands.py tests/sdk/test_sdk.py -q
110 passed, 1 warning
```

Earlier broader focused run after integration:

```text
160 passed, 1 warning
```

No-rent/preflight live validation:

- `vastai --raw self-test machine 0`: structured `no_offer`, no rental.
- `vastai --raw self-test machine 54894`: direct ports, download, and upload failures show actual values, thresholds, purpose, and remediation; no rental.
- `vastai --raw self-test machine 16689`: GPU RAM failure shows `5.99 GiB`, required `> 7 GiB`; no rental.
- `vastai --raw self-test machine 68277`: PCIe/system RAM/CPU failures show actual values and thresholds; no rental.

Paid runtime red validation:

- Machine `19028`, dogfood image `jjziets/vast-self-test:con1510-cuda-13.0-amd64`, instance `38084614`: failed before runtime start with OCI/CDI GPU injection error; destroyed.
- Machine `23973`, same dogfood image, instance `38085349`: reproduced OCI/CDI startup failure; destroyed.
- Machine `19028`, official image `vastai/test:self-test-cuda-13.0`, instance `38094685`: reproduced the same OCI/CDI failure and the CLI classified it as `daemon_startup_failed`; destroyed.

Paid green validation with structured self-test image:

- Machine `55752`, instance `38090857`: strict one-GPU pass; `/summary.json` confirmed `gpu_count: 1`; all stages passed; destroyed.
- Machine `46001`, instance `38092686`: strict one-GPU pass after removing `jupyter_direc`; instance used `/ssh`, `5000/tcp` stayed mapped, `/summary.json` confirmed `gpu_count: 1`; all stages passed; destroyed.
- Machine `21044`, instance `38089696`: all stages passed, but selected a 2-GPU offer from the machine despite starting from a 1-GPU candidate search. This exposed the existing machine-level offer-selection caveat.

Every paid run ended with a final `show instances` audit returning zero active instances.

## Related Work

- `vast-ai/self-test#1` restores the missing self-test runtime source files (`remote.py`, `remote.sh`) and validates the current published image family.
- P3 structured image progress is implemented on `jjziets/self-test`, branch `CON-1510-structured-self-test-progress`, and was dogfooded via `jjziets/vast-self-test:con1510-cuda-13.0-amd64`.
- P4 docs/UI extraction is intentionally deferred until the diagnostic schema stabilizes.

### DIFF
--- a/vastai/cli/self_test/machine_diagnostics.py
+++ b/vastai/cli/self_test/machine_diagnostics.py
@@ -1,0 +1,287 @@
+"""Structured diagnostics for ``vastai self-test machine``."""
+
+from vastai.cli.util import required_inet_mbps
+
+
+DIAGNOSTICS_VERSION = "1"
+
+
+def safe_float(value):
+    if value is None:
+        return 0.0
+    try:
+        return float(value)
+    except (ValueError, TypeError):
+        return 0.0
+
+
+def gpu_ram_gib(value):
+    """Normalize API GPU RAM values for display.
+
+    Search offer output historically displays `gpu_ram` after dividing by
+    1000, while tests and some call paths use a direct GiB-style value. Treat
+    large values as MiB so host-facing self-test output does not print
+    impossible values such as `32768 GB`.
+    """
+    raw = safe_float(value)
+    if raw > 256:
+        return raw / 1024
+    return raw
+
+
+def compact_offer_metadata(offer):
+    if not offer:
+        return None
+    fields = (
+        "id",
+        "machine_id",
+        "gpu_name",
+        "num_gpus",
+        "dph_total",
+        "dlperf",
+        "cuda_max_good",
+        "compute_cap",
+        "reliability",
+        "direct_port_count",
+        "pcie_bw",
+        "inet_down",
+        "inet_up",
+        "gpu_ram",
+        "gpu_total_ram",
+        "cpu_ram",
+        "cpu_cores",
+        "rentable",
+        "rented",
+    )
+    return {field: offer.get(field) for field in fields if field in offer}
+
+
+def base_result(machine_id):
+    return {
+        "success": False,
+        "reason": "",
+        "warning": None,
+        "phase": "preflight",
+        "diagnostics_version": DIAGNOSTICS_VERSION,
+        "machine_id": machine_id,
+        "offer": None,
+        "checks": [],
+        "failure": None,
+        "failure_code": None,
+        "stage": None,
+        "error": None,
+        "diagnostics": {},
+    }
+
+
+def _check(
+    check_id,
+    title,
+    actual,
+    required,
+    operator,
+    unit,
+    passed,
+    purpose,
+    remediation,
+):
+    status = "pass" if passed else "fail"
+    return {
+        "id": check_id,
+        "title": title,
+        "status": status,
+        "actual": actual,
+        "required": required,
+        "operator": operator,
+        "unit": unit,
+        "summary": f"{title}: actual {actual} {unit}, required {operator} {required} {unit}",
+        "purpose": purpose,
+        "remediation": remediation,
+    }
+
+
+def preflight_requirement_checks(offer):
+    gpu_total_ram = safe_float(offer.get("gpu_total_ram"))
+    per_gpu_ram_gib = gpu_ram_gib(offer.get("gpu_ram"))
+    required_mbps = required_inet_mbps(gpu_total_ram)
+    cpu_ram = safe_float(offer.get("cpu_ram"))
+    cpu_cores = int(safe_float(offer.get("cpu_cores")))
+    num_gpus = int(safe_float(offer.get("num_gpus")))
+    required_cpu_ram = 0.95 * gpu_total_ram
+    required_cpu_cores = 2 * num_gpus
+
+    return [
+        _check(
+            "cuda.version",
+            "CUDA version",
+            safe_float(offer.get("cuda_max_good")),
+            11.8,
+            ">=",
+            "CUDA",
+            safe_float(offer.get("cuda_max_good")) >= 11.8,
+            "The self-test image needs a host driver stack compatible with CUDA 11.8 or newer.",
+            "Update the NVIDIA driver/CUDA stack, then confirm with: vastai search offers "
+            f"'machine_id={offer.get('machine_id') or 'unknown'} rentable=any rented=any'",
+        ),
+        _check(
+            "reliability",
+            "Reliability",
+            safe_float(offer.get("reliability")),
+            0.90,
+            ">",
+            "ratio",
+            safe_float(offer.get("reliability")) > 0.90,
+            "Self-test uses reliability as a guardrail before launching a temporary instance.",
+            "Let the host stabilize and resolve recent failures before retrying the self-test.",
+        ),
+        _check(
+            "network.direct_ports",
+            "Direct port count",
+            safe_float(offer.get("direct_port_count")),
+            3,
+            ">",
+            "ports",
+            safe_float(offer.get("direct_port_count")) > 3,
+            "The tester needs enough directly mapped ports for remote progress and SSH checks.",
+            "Open additional direct ports or adjust host firewall/NAT settings, then rerun verification.",
+        ),
+        _check(
+            "pcie.bandwidth",
+            "PCIe bandwidth",
+            safe_float(offer.get("pcie_bw")),
+            2.85,
+            ">",
+            "GB/s",
+            safe_float(offer.get("pcie_bw")) > 2.85,
+            "Low PCIe bandwidth can make GPU stress and transfer checks fail or time out.",
+            "Check BIOS PCIe generation/lane settings and confirm GPUs are seated in full-speed slots.",
+        ),
+        _check(
+            "network.download",
+            "Download speed",
+            safe_float(offer.get("inet_down")),
+            round(required_mbps, 2),
+            ">=",
+            "Mb/s",
+            safe_float(offer.get("inet_down")) >= required_mbps,
+            "The bandwidth floor scales with total VRAM so large GPU hosts can complete data movement tests.",
+            "Improve host download bandwidth or reduce contention, then rerun the Vast host verification.",
+        ),
+        _check(
+            "network.upload",
+            "Upload speed",
+            safe_float(offer.get("inet_up")),
+            round(required_mbps, 2),
+            ">=",
+            "Mb/s",
+            safe_float(offer.get("inet_up")) >= required_mbps,
+            "The tester needs enough upload bandwidth to report progress and complete network checks.",
+            "Improve host upload bandwidth or reduce contention, then rerun the Vast host verification.",
+        ),
+        _check(
+            "gpu.ram",
+            "GPU RAM",
+            round(per_gpu_ram_gib, 2),
+            7,
+            ">",
+            "GiB",
+            per_gpu_ram_gib > 7,
+            "The verification workload requires more than 7 GB of VRAM per GPU.",
+            "Use a GPU with more VRAM for this self-test.",
+        ),
+        _check(
+            "system.ram",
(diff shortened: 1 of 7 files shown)


robballantyne (approved):
Reviewing on the assumption #405 and #407 are already merged. Subtracting their changes, the net 408-only delta is roughly:

  - New vastai/cli/self_test/ package with two pure modules (machine_diagnostics.py ~287 lines, runtime_diagnostics.py ~430 lines)
  - Substantial refactor of self_test__machine in vastai/cli/commands/machines.py
  - Schema-versioned structured output (diagnostics_version, checks, failure, failure_code, phase, stage, error, offer, diagnostics)
  - --test-image flag + VAST_SELF_TEST_IMAGE env override
  - Runtype change: "jupyter_direc ssh_direc ssh_proxy" → "ssh_direc ssh_proxy" (drops Jupyter for self-test)
  - API key redaction in error output
  - Runtime failure classification (classify_status_msg, LegacyProgressParser, 22-code failure catalog)
  - 404 handling added to final cleanup HTTPError branch
  - Raw output mechanism changes from inline print(json.dumps); sys.exit(0) to return result (now handled by main.py:138-145)

  Overview

  This is a substantial diagnostics rework. The old self-test CLI printed inconsistent free-text errors, duplicated offer-search logic across
   check_requirements and search_offers_and_get_top, and produced raw JSON output that downstream tools couldn't reliably classify. PR #408
  introduces a structured failure schema, a stable failure-code catalog, and a clean separation between pure diagnostic logic and live API
  orchestration. It also addresses several real footguns: literal {machine_id} placeholders leaking into user-facing remediation strings,
  stale --filter syntax in hints, double offer searches that could disagree, and unredacted API keys in error messages.

  What's good

  - Pure modules in vastai/cli/self_test/. machine_diagnostics.py and runtime_diagnostics.py have no Vast API dependencies and no I/O — every
   function takes data and returns data. This is exactly the right separation for diagnostics; tests can hit them directly without mocking
  HTTP.
  - Schema versioning (diagnostics_version: "1"). Anyone consuming --raw output now has a stable contract.
  - Failure catalog (22 codes, single source of truth in FAILURE_CATALOG). Each code has a summary, remediation, and suggested_steps. This is
   the right shape for downstream tools and dashboards to surface remediation without re-parsing free text.
  - Reusing one offer across preflight and rental (selected_offer_for_self_test). Previously, check_requirements selected a top offer for
  validation, then search_offers_and_get_top re-selected — they could disagree. Now there's one source of truth.
  - Two-tier offer probe (strict_query → broader_query on miss). The user gets a meaningful diagnostic distinguishing "no offer at all" from
  "exists but not rentable right now" — emitted as no_offer vs no_rentable_offer.
  - API key redaction (safe_error). re.sub(r"([?&]api_key=)[^&\s]+", r"\1REDACTED", ...) is applied at every print(... {error}) and
  debug_print(... {error}) site. Important security fix — previously the full URL with the key would land in logs and self-test JSON.
  - Replaced stale literal {machine_id} placeholders with f-strings, and updated --filter syntax to current positional 'machine_id=X 
  rentable=any rented=any'. Real user-facing improvement.
  - --test-image + VAST_SELF_TEST_IMAGE for dogfooding new self-test images before retagging the official vastai/test:* family. Tracked in
  raw output as diagnostics.image.override: true.
  - 404 in cleanup is no longer a leak warning — the new HTTPError handler distinguishes "instance gone" from "destroy genuinely failed."

  Issues / suggestions

  Concerns worth addressing

  - GPU RAM unit-normalization heuristic (machine_diagnostics.py:gpu_ram_gib):
  if raw > 256:
      return raw / 1024
  return raw
  - The threshold is "no GPU has more than 256 GB so anything bigger must be in MiB." Today's largest GPU (B200) is 192 GB, so this works for
   now, but it's a hidden time bomb when new SKUs ship. Better to either (a) bump to a much larger threshold (4096+), (b) trust the API
  value's documented unit and convert explicitly, or (c) check whether the offer dict has both gpu_ram and gpu_total_ram and use their ratio
  against num_gpus as a sanity check.
  - safe_error() only redacts api_key= in URL query strings. It won't catch:
    - Authorization: Bearer <token> headers if those ever appear in error messages
    - ?token=, ?key=, or other parameter names
    - JSON bodies with "api_key": "..."

  Probably sufficient for current Vast API URL shapes, but a comment saying "intentionally narrow; only redacts the URL form we currently
  emit" would document the limitation. If you want defense in depth, also redact (Bearer\s+)[\w.-]+ and ("api_key"\s*:\s*")[^"]+.
  - 404 cleanup branch is partly redundant with #405. Once #405 lands, the info = show_instance(...) call already returns None for the
  deleted case and the if not info: ... info = {} block handles it. The new explicit except requests.exceptions.HTTPError as e: if 
  e.response.status_code == 404: then catches a different failure mode (where the API call returns 404 directly rather than {"instances": 
  null}). Both paths can occur, so this is reasonable defense-in-depth, but a one-line comment explaining "404 here means the destroy itself
  observed the instance as already gone, distinct from show_instance returning None earlier" would help.

  Smaller nits

  - failure_catalog() returns a fresh dict every call (good — JSON-serializable, mutation-safe), but it allocates ~22 dicts each invocation.
  Not perf-critical, but if it's called per check, consider memoizing.
  - compact_offer_metadata field list is hardcoded. A new offer field added to the API won't appear in self-test output until someone updates
   this list. Probably intentional (you want a stable shape for downstream consumers), but worth a comment saying so.
  - STAGE_STARTUP, STAGE_NCCL, etc. as module-level constants vs the Stage enum pattern Python typically uses. Constants are fine for a
  closed set; just noting it'd be more discoverable as class Stage(StrEnum).

  Legacy vast.py

  Same observation as my reviews of #405 and #407 — the legacy monolithic vast.py:8363 still has the old self_test__machine and won't gain
  any of these diagnostics. Since the project documents vast.py as an alternative install method, users on that path won't see structured
  failures, failure codes, API key redaction, or the offer-reuse fix. Given the size of this refactor, mirroring it into vast.py may not be
  practical — but worth a project decision (and perhaps a deprecation note in vast.py) rather than silent divergence.

  Risk

  Moderate. The new pure modules are well-isolated and low-risk. The self_test__machine refactor is sizable but well-covered by tests, and
  the paid green-validation runs on three machines give real confidence. One thing to watch post-merge: the GPU RAM normalization heuristic
  has a future failure mode at GPUs > 256 GB.

  Verdict

  Strong work. Real ergonomic and structural improvements, with the right separation between pure logic and orchestration. Before merge I'd
  ask for: bump or comment the gpu_ram_gib threshold so its hidden-future-failure isn't silent.

  The legacy vast.py divergence is worth a project-level decision but shouldn't block this PR.

Live tested with both current images and provided test image. Looks good

jjziets:
Docker Hub readiness update: the production tag required by this PR now exists.

- `vastai/test:self-test-cuda-13.3`
- digest: `sha256:da1098d6adf9eba5fef326eaa9871edf916b4ab150290291eed5521944e2ab5e`
- platforms: `linux/amd64`, `linux/arm64`

This was created from the safe CON-1510 tag `vastai/test:self-test-con1510-cuda-13.3`. Existing production tags for 11.8, 12.8, and 13.0 were not overwritten.

jjziets:
Compatibility update: #408 now routes new CLI clients to a non-overwriting image family.

New #408 default image family:

- `vastai/test:self-test-v2-cuda-11.8`
- `vastai/test:self-test-v2-cuda-12.8`
- `vastai/test:self-test-v2-cuda-13.0`
- `vastai/test:self-test-v2-cuda-13.3`

These tags were created from the rebuilt CON-1510 image family. Existing production tags used by older CLIs (`self-test-cuda-11.8`, `self-test-cuda-12.8`, `self-test-cuda-13.0`) were not overwritten, so older CLI users keep their current image behavior. Focused verification after the routing change: `167 passed, 1 warning`.

vastai/cli/self_test/machine_diagnostics.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Search offer output historically displays `gpu_ram` after dividing by
+    1000, while tests and some call paths use a direct GiB-style value. Treat
+    large values as MiB so host-facing self-test output does not print
+    impossible values such as `32768 GB`.
+    """
+    raw = safe_float(value)
+    if raw > 256:
+        return raw / 1024
+    return raw
```
robballantyne: 1 x B300 will show 0.28125 vram

jjziets:
Addressed Rob's B300/high-VRAM review point in commit 70ebc06.\n\nWhat changed:\n- Removed the `>256 => divide by 1024` VRAM heuristic.\n- Self-test preflight now derives per-GPU VRAM from canonical fields first: `gpu_total_ram / num_gpus / 1024`.\n- `gpu_ram` is only a fallback when canonical fields are missing/invalid.\n\nCoverage added:\n- B300-style regression: `gpu_ram=288`, `gpu_total_ram=288*1024`, `num_gpus=1` reports `288 GiB`, not `0.28125 GiB`.\n- Multi-GPU total-VRAM derivation regression.\n- Fallback regression for `gpu_ram=6144` -> `6.0 GiB` when `gpu_total_ram` is unavailable.\n\nLocal validation passed:\n- `tests/cli/test_machines_commands.py tests/cli/test_runtime_diagnostics.py`: 45 passed\n- broader CLI/SDK subset: 164 passed\n\nPaid dogfood after the fix rented/attempted live hosts and confirmed the rebased #408 CLI selects the v2 image family and cleans up instances. The available live hosts hit host-side Docker/CDI/pull failures rather than a self-test pass, all classified by the new diagnostics and all destroyed successfully. Final active-instance audit was zero.

jjziets:
Additional live B300 validation for the VRAM fix:\n\nThe marketplace name is `B300 SXM6 AC`, and live offers report `gpu_ram=275040` and `gpu_total_ram=275040` per GPU. I ran raw self-test preflight only, without `--ignore-requirements`, on expensive B300 candidates that were expected to fail before rental. No runtime rentals were made.\n\nResults:\n- machine `136638`: selected 8x B300 offer, `gpu.ram actual=268.59 GiB`, status `pass`; failed PCIe/network preflight.\n- machine `136640`: selected 8x B300 offer, `gpu.ram actual=268.59 GiB`, status `pass`; failed PCIe/upload preflight.\n- machine `123158`: selected 8x B300 offer, `gpu.ram actual=268.59 GiB`, status `pass`; failed reliability/system RAM preflight.\n- machine `134965`: selected 8x B300 offer, `gpu.ram actual=268.59 GiB`, status `pass`; failed download/upload preflight.\n\nFinal active-instance audit returned `[]`.

PR #409, opened 2026-05-29
CON-1514: Disambiguate self-test common failures

jjziets:
## Summary

Implements CLI-side CON-1514 diagnostics on top of the #408 self-test diagnostics work.

This PR improves two commonly reported self-test failure classes:

- "Machine ID not found or not rentable"
- "No response for 120s"

It also addresses the dogfood finding from the `--ignore-requirements` flow:

- If minimum requirements fail but `--ignore-requirements` is set, those failures are preserved as skipped-gate metadata.
- If the runtime checks pass, the top-level result is `success=true` with no contradictory top-level `failure_code`.
- Output now communicates: runtime checks passed, minimum requirement checks were ignored, and the run does not qualify the machine for verification.

## What Changed

- Adds root-state inference for no-offer/no-rentable cases:
  - `currently_rented`
  - `deverified_or_below_threshold`
  - `api_permission_failed`
  - `zero_active_offers`
  - `offline_or_not_listed`
  - `unknown_no_rentable_offer`
- Adds `root_state`, `confidence`, `evidence`, and `suggested_steps` to structured no-offer failures.
- Preserves more offer status metadata, including `vericode`, `verified`, `verification`, and `error_description`.
- Converts offer/machine lookup 401/403 failures into `api_permission_failed` instead of generic `unexpected_error`.
- Adds structured progress endpoint diagnostics:
  - attempted URL
  - public IP
  - container port
  - host port
  - timeout
  - attempt count
  - first-connection state
  - last error/status
  - mapped ports
- Makes `progress_endpoint_unreachable`, `progress_endpoint_lost`, `progress_empty_timeout`, and `progress_port_not_mapped` easier to diagnose.
- Keeps API key query values redacted in host-visible diagnostics.

## Relationship To #408 And Images

This PR is stacked on #408 while #408 is still open.

The image side must move with the CLI self-test changes:

- New CLI flows use the non-breaking `vastai/test:self-test-v2-cuda-*` image family introduced with #408.
- Existing legacy tags are not overwritten, so older CLIs keep using the old image family.
- No additional image rebuild is required specifically for CON-1514; this PR consumes the v2 structured progress behavior already prepared in the self-test image repo.
- If this PR supersedes #408 before broad use, that is fine: #409 effectively carries the #408 image routing plus the CON-1514 diagnostic refinements.

## Validation

Mocked/local:

- Focused self-test diagnostics:
  - `53 passed, 1 warning`
- Broader required suite:
  - `172 passed, 1 warning`
[...]

PR #410, opened 2026-06-02
CON-1519: add self-test diagnostic support bundles

jjziets:
## Summary

Implements CON-1519 support-bundle handoff for host self-test failures.

This PR adds:
- automatic diagnostic bundle creation on `vastai self-test machine <machine_id>` failure;
- clear console output showing the tarball path and bundle contents before sharing;
- `vastai dump-logs <machine_id>` for on-demand bundle creation;
- redaction for API keys, tokens, secrets, authorization fields, `instance_api_key`, `jupyter_token`, and `host_port_range` style values;
- per-artifact truncation to keep bundles small;
- raw diagnostics support via `result["diagnostics"]["support_bundle"]`;
- `--support-bundle-dir` and `--no-support-bundle` controls;
- `VAST_SELF_TEST_SUPPORT_BUNDLE=0` escape hatch for tests/automation.

## Bundle Contents

The bundle includes the self-test output/result plus best-effort local host artifacts:
- `/var/lib/vastai_kaalia/kaalia.log*` rotations, truncated/redacted;
- filtered `dmesg -T` output;
- `journalctl -b -p err..alert -u docker -u vastai_kaalia --since -24h`;
- `/etc/docker/daemon.json`;
- `nvidia-smi` and `nvidia-smi -L`;
- `docker info`;
- `/proc/mounts` entries for `/var/lib/docker`;
- `ip addr`;
- summarized `ss -tlnp` output without the process column.

Unavailable artifacts are recorded in `collection-errors.json` and shown in the console summary.

## Validation

Local tests:
- `uv run --with pycryptodome --with pytest --with pytest-mock python -m pytest tests/cli/test_self_test_support_bundle.py tests/cli/test_machines_commands.py -q`
  - 51 passed
- `uv run --with pycryptodome --with pytest --with pytest-mock python -m pytest tests/cli -q`
  - 295 passed
- `git diff --check`
  - clean

Local smoke:
- `vastai dump-logs 42 --output-dir <tmp>` created a tarball, printed the contents, and the tarball opened successfully.
- `vastai self-test machine 56732 --support-bundle-dir <tmp>` failed in preflight with no rental, automatically wrote a bundle, printed the bundle path/contents, and the tarball opened successfully.

## Stack / Merge Note

This is stacked on #409 because CON-1519 builds on the current self-test diagnostics branch. Merge/rebase this after #409 lands.

### DIFF
--- a/tests/cli/test_self_test_support_bundle.py
+++ b/tests/cli/test_self_test_support_bundle.py
@@ -1,0 +1,118 @@
+"""Tests for self-test diagnostic support bundles."""
+
+import json
+import tarfile
+from unittest.mock import Mock
+
+import pytest
+
+from vastai.cli.self_test.support_bundle import create_support_bundle
+
+
+def test_support_bundle_contains_redacted_result_and_cli_output(tmp_path):
+    bundle = create_support_bundle(
+        machine_id="42",
+        output_dir=str(tmp_path),
+        result={
+            "machine_id": "42",
+            "error": "request failed: https://console.vast.ai/?api_key=secret",
+            "diagnostics": {"jupyter_token": "secret-token"},
+        },
+        cli_output=["Starting self-test", "token=supersecret"],
+        run_started_at="20260602T100000Z",
+        command=["vastai", "--api-key", "secret", "self-test", "machine", "42"],
+        secrets=["secret", "supersecret"],
+        include_host_logs=False,
+    )
+
+    with tarfile.open(bundle["path"], "r:gz") as tar:
+        names = set(tar.getnames())
+        result_json = tar.extractfile("self-test-result.json").read().decode()
+        output_log = tar.extractfile("self-test-output.log").read().decode()
+        manifest = json.loads(tar.extractfile("manifest.json").read().decode())
+
+    assert {"manifest.json", "self-test-result.json", "self-test-output.log", "collection-errors.json"} <= names
+    assert "secret" not in result_json
+    assert "secret" not in output_log
+    assert "REDACTED" in result_json
+    assert "REDACTED" in output_log
+    assert manifest["machine_id"] == "42"
+    assert manifest["run_started_at_utc"] == "20260602T100000Z"
+
+
+def test_self_test_failure_creates_support_bundle(
+    parse_argv, patch_get_client, monkeypatch, tmp_path, capsys
+):
+    from vastai.cli.commands import machines
+
+    def fake_bundle(**kwargs):
+        assert kwargs["machine_id"] == "42"
+        assert kwargs["output_dir"] == str(tmp_path)
+        assert kwargs["include_host_logs"] is True
+        assert kwargs["result"]["failure_code"] == "preflight_requirements_failed"
+        assert any("Preflight diagnostics for machine 42 failed:" in line for line in kwargs["cli_output"])
+        return {
+            "path": str(tmp_path / "vast_selftest_42_20260602T100000Z.tar.gz"),
+            "created_at_utc": "20260602T100000Z",
+            "size_bytes": 123,
+            "files": ["manifest.json", "self-test-result.json", "self-test-output.log"],
+            "collection_errors": [],
+        }
+
+    bad_offer = {
+        "id": 777,
+        "cuda_max_good": "11.7",
+        "compute_cap": 750,
+        "dlperf": 1,
+        "reliability": 0.99,
+        "direct_port_count": 1,
+        "pcie_bw": 1.0,
+        "gpu_total_ram": 6 * 1024,
+        "inet_down": 10,
+        "inet_up": 10,
+        "gpu_ram": 6 * 1024,
+        "cpu_ram": 1024,
+        "cpu_cores": 1,
+        "num_gpus": 1,
+        "machine_id": 42,
+    }
+    monkeypatch.setenv("VAST_SELF_TEST_SUPPORT_BUNDLE", "1")
+    monkeypatch.setattr(machines, "create_support_bundle", fake_bundle)
+    monkeypatch.setattr(machines.offers_api, "search_offers", Mock(return_value=[bad_offer]))
+
+    args = parse_argv(["self-test", "machine", "42", "--support-bundle-dir", str(tmp_path)])
+    with pytest.raises(SystemExit) as exc_info:
+        args.func(args)
+
+    captured = capsys.readouterr()
+    assert exc_info.value.code == 1
+    assert "Self-test diagnostic bundle saved to:" in captured.out
+    assert "self-test-result.json" in captured.out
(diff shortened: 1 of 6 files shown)


jjziets:
Update after reviewing the host/laptop boundary for CON-1519:

- Corrected the automatic self-test failure bundle so it no longer assumes the CLI is running on the Vast host. By default it now contains CLI-visible evidence only: self-test output/result, manifest, collection errors, and Vast API instance logs when a self-test instance exists.
- Added instance log collection through the existing Vast logs API before failure cleanup destroys the rented self-test instance:
  - `instance/container.log`
  - `instance/daemon.log`
  - `instance/show-instance.json`
- Updated `vastai dump-logs <machine_id>` to be honest about scope. It can pull the same API-visible instance logs with `--instance-id <id>`. Local host OS/kaalia artifacts are now explicit via `--include-local-host-artifacts`, with help text saying to use it only when running on the actual host machine.
- This means laptop-run self-tests no longer package laptop `dmesg`/journal/kaalia-looking paths as if they came from the host. Full host-side kaalia/dmesg/journal/daemon.json collection still needs daemon/backend support as a follow-up feature.

Validation:
- `uv run --with pycryptodome --with pytest pytest tests/cli/test_self_test_support_bundle.py -q` -> 5 passed
- `uv run --with pycryptodome --with pytest pytest tests/cli -q` -> 297 passed
- `git diff --check` -> clean

robballantyne (changes requested):
Reviewed the #410-specific delta on top of #409. The support-bundle design is solid and security-conscious (layered redaction, host artifacts opt-in only on the auto-failure path, `ss` process column stripped, per-artifact timeouts/guards, good test coverage). Two items before merge:

## 🔴 Blocker — CI is red on Windows

`test_self_test_runtime_failure_bundle_includes_instance_logs` fails on `windows-latest` (`assert 123 in set()` — the instance is never destroyed).

Root cause: the CUDA-mapping reason strings use `→` (U+2192) (machines.py ~1016, 1049–1055, inherited from #409). Every prior self-test test used `--raw`, which suppresses `progress_print`, so the char was never written to a console. This PR's new test runs **without `--raw`**, so `progress_print` writes `→` to stdout and Windows' cp1252 console raises `UnicodeEncodeError: 'charmap' codec can't encode character '→'`, aborting the create/destroy flow.

Fix: replace `→` with `->` in those reason strings (simplest), or force UTF-8 stdout (`sys.stdout.reconfigure(encoding="utf-8", errors="replace")`). The ASCII fix belongs in this PR since the failing test is here.

## 🟠 Robustness — bundle creation can mask the real failure

`ensure_support_bundle()` → `create_support_bundle()` is not wrapped in try/except. `mkdir`, `tarfile.open`, and `stat` can all raise (read-only/missing `--support-bundle-dir`, disk full, permissions). Since this runs inside `finish_failure()` and again in the final non-raw block, a write error replaces the clean `Test failed: …` exit with a traceback — best-effort diagnostics shouldn't crash the command they're meant to help. Please wrap in `try/except Exception`, warn, and continue to the original failure exit. Same for `dump_logs`.

## 🟡 Minor (non-blocking)

- **Ineffective tar traversal guard:** `safe_name = re.sub(r"(^/|\.\.|\\)", "_", name)` does not actually match `..` (verified `../../evil` passes through unchanged). Low severity — names are developer-controlled and the CLI never extracts the tar — but if it's meant as a guard, fix to e.g. `r"(^/+|\.\.|\\)"`.
- **JSON truncation:** `manifest.json`/`self-test-result.json` are re-truncated to `MAX_LOG_BYTES` in the tar loop; a >256 KB result would be cut mid-JSON. Consider exempting JSON artifacts.
- **File perms:** bundle is written to `/tmp` with default umask (world-readable on shared hosts). Contents are redacted, but `chmod 0o600` is cheap defense-in-depth.
- **Nit:** `dump-logs` `usage=` string omits `--include-local-host-artifacts`.

Keep the stated merge order (land after #409).

jjziets:
Addressed the review items in the latest commits.

Changes made:
- Replaced the Unicode arrows in the self-test CUDA image selection messages with ASCII `->`, fixing the Windows `cp1252` stdout failure exposed by the new non-raw test.
- Wrapped `create_support_bundle()` in best-effort error handling for both auto self-test failure bundles and `dump-logs`, so bundle write errors warn but do not mask the original self-test failure.
- Fixed the tar archive-name sanitization so `..` is actually removed from archive paths.
- Exempted JSON artifacts from the final log truncation pass so `manifest.json` / `self-test-result.json` are not cut mid-JSON.
- Set created bundle permissions to `0600`.
- Updated `dump-logs` usage text to include `--include-local-host-artifacts`.
- Also fixed the install-smoke failure where `vastai --help` could fail if the live GPU-name lookup returned 403; the CLI now falls back to unrestricted GPU-name choices instead of failing during command registration.

Validation:
- Local focused tests: `tests/cli/test_self_test_support_bundle.py tests/cli/test_util.py` -> 60 passed
- Local CLI suite: `tests/cli` -> 302 passed
- Local `vastai --help` smoke passed
- Fresh GitHub Actions run passed on all matrix jobs, including:
  - `unit-and-integration (windows-latest)` passed
  - `install-smoke (windows-latest)` passed
  - Ubuntu/macOS unit + smoke jobs passed

Run: https://github.com/vast-ai/vast-cli/actions/runs/26816420547

robballantyne (approved):
Approving — all requested changes addressed in f4cc2dd1, and verified. CI is now green across all platforms including Windows.

- **Windows blocker**: `→` replaced with ASCII `->` in the CUDA-mapping reason strings; the non-raw runtime-failure test now passes on windows-latest.
- **Bundle robustness**: `create_support_bundle` is wrapped in try/except in both paths. In self-test, a failure records `diagnostics.support_bundle_error`, warns, and returns None so the original failure exit still proceeds; `dump-logs` warns + exits 1 (or returns an error dict in --raw). Diagnostics can no longer mask the real failure.
- **Tar traversal guard**: `_safe_archive_name` now uses `r"(^/+|\.\.|\\)"` — verified `../../evil -> _/_/evil`, leading slashes and backslashes neutralised, legit names preserved.
- **JSON truncation**: `*.json` artifacts are exempted from byte-tail truncation via `JSON_ARTIFACT_RE`.
- **File perms**: bundle written then `chmod 0o600`.
- **Nit**: `dump-logs` usage string now lists `--include-local-host-artifacts`.

Also reviewed the bonus commit 979bfd33 (CLI help robust without GPU name lookup): `_get_gpu_names()` now returns None on network/cache/parse failure instead of raising. Since it's evaluated at import time as `choices=_get_gpu_names()`, this prevents a network blip from breaking the entire CLI; `choices=None` is argparse's no-restriction default, so the degradation is safe. Good catch.

Land after #409 per the stated merge order.

PR #417, opened 2026-06-10
cache benchmark runs

vastzuby:
**overview**

`vastai run benchmarks` rents a real GPU, measures perf, and writes a record to the benchmarks table. 
  
this PR will use that table as a cache so that we dont have run duplicate benchmarks. 
a benchmark is duplicate if it has the same gpu type, num_gpus, template, and ran in past 30 days
  
**caching**

cache hit -> skip the rental and return the median of the reported perfs.
cache miss -> run a new benchmark 
  
note that the cache reports the perf while a real benchmark will return the perf/$ but we dont know the gpu price yet for historical benchmarks.

**edge cases**

if no offers available,  we flag to user  ("no offers available to rent right now") 

**future work**

update the benchmarks table to track the price of a benchmarked gpu.
that way, we can return the perf/$ for cached responses.

also includes unrelated test fix where we refactored env -> current in perform update() but test file was still asserting the old layout  (env/bin/vastai)

**unrelated test fix**
The XDG-layout refactor (#452) renamed  env/ → current/ and but test_prune_failure_does_not_fail_the_update (added in #447) still asserted the old env/ path

### DIFF
--- a/vastai/cli/commands/benchmarks.py
+++ b/vastai/cli/commands/benchmarks.py
@@ -48,20 +48,15 @@ DEFAULT_GPUS = {
 # One worker per GPU, kept alive while idle so we can read its measured perf.
 ENDPOINT_CONFIG = {"cold_workers": 1, "max_workers": 1, "min_load": 1.0}
 
-# Timing knobs, in the order they fire during a run.
-RENTAL_TIMEOUT = 120                 # give up if the autoscaler hasn't rented a worker by now
-DEFAULT_BENCHMARK_TIMEOUT = 30 * 60  # default --timeout: max wait for a measured result
-WORKER_POLL_INTERVAL = 10.0          # base seconds between worker-status polls (jitter added at call site)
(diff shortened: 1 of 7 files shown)


vastai/cli/commands/benchmarks.py
```diff
@@ -291,39 +252,79 @@ def _get_gpu_chunk_size(n):
...
+
+
+def lookup_cached_benchmark(vast, *, gpu_name, num_gpus, template_hash,
+                             template_id, max_age_days):
+    """Recent reported benchmarks for this exact spec (median + spread), or None.
+
+    Template is matched client-side: rows carry template_hash or template_id
+    depending on how the benchmarked workergroup was created.
+    """
+    query = {
```
LucasArmandVast: We should include a limit on this query, since it could potentially return 1000s of rows for common template/num_gpu/gpu_name combos (i.e. ComfyUI on 1x5090)

LucasArmandVast (approved):

PR #418, opened 2026-06-11
fix(AUTO-959): shared util moved to set up for GA4 analytics tracking for created_endpoint through the sdk flow

madisonvast:
## What

All HTTP clients in the package now send an identifying `User-Agent` header on every request to the console API:

- SDK surfaces (`VastAI`, serverless client, `SyncClient`/`AsyncClient`): `vastai-sdk/<version>`
- CLI (`vastai` command): `vastai-cli/<version>`

## Why

The webserver emits a GA4 `created_endpoint` event when an endpoint is created, with a
`source` param telling us where it came from. Detection was cookie-based ("no cookies = CLI"),
so SDK, CLI, and curl traffic were indistinguishable — everything reported as `CLI`, and there
was no way to track SDK adoption. This pairs with a server-side change (AUTO-959) that reads
these UA prefixes to label `source` as `SDK` / `CLI` / `UI`.

Versioned UAs also let us slice analytics by client version later (e.g. measuring upgrade lag).

## How

- `VastClient` gains a `client_type` kwarg (default `"sdk"`); the CLI's `get_client()` passes
  `"cli"`. Needed because both layers share the same HTTP client.
- The serverless async client (`_make_request`) and `_BaseClient` (sync/async instance clients)
  send the SDK UA unconditionally — they have no CLI entry point.
- `_get_git_version` / `_get_local_version` / `VERSION` moved from `vastai/cli/util.py` to
  `vastai/utils.py` so the API layer can import the version without a circular import
  (`cli/util.py` imports from `api/client.py`). `cli/util.py` re-exports them, so `--version`
  and existing imports are unchanged.

## Compatibility

- No behavior change for any API call — requests just carry one extra header.
- Old SDK/CLI versions in the wild keep being labeled `CLI` server-side via the existing
  cookie heuristic; labels improve as clients upgrade.

## Testing

- Full unit suite passes (381 passed).
- Replaced `test_empty_when_no_key` (headers are never empty now) with assertions for the
  no-auth case and both `client_type` UA variants.
- Verified against a local webserver: endpoint creation with the new headers logs GA4 events
  with `source: "SDK"` / `source: "CLI"` respectively.

### DIFF
--- a/tests/api/test_client.py
+++ b/tests/api/test_client.py
@@ -5,6 +5,7 @@ import pytest
 import requests
 from unittest.mock import patch, MagicMock
 from vastai.api.client import VastClient
+from vastai.utils import VERSION
 
 
 class TestBuildUrl:
@@ -57,10 +58,20 @@ class TestBuildHeaders:
         h = c._build_headers()
         assert h["Authorization"] == "Bearer mykey"
 
-    def test_empty_when_no_key(self):
+    def test_no_auth_when_no_key(self):
         c = VastClient(api_key=None)
         h = c._build_headers()
-        assert h == {}
+        assert "Authorization" not in h
+
+    def test_user_agent_defaults_to_sdk(self):
+        c = VastClient(api_key=None)
+        h = c._build_headers()
+        assert h["User-Agent"] == f"vastai-sdk/{VERSION}"
+
+    def test_user_agent_client_type(self):
+        c = VastClient(api_key=None, client_type="cli")
+        h = c._build_headers()
+        assert h["User-Agent"] == f"vastai-cli/{VERSION}"
 
 
 class TestHttpMethods:
--- a/vastai/_base.py
+++ b/vastai/_base.py
@@ -1,6 +1,8 @@
 import os
 from typing import Optional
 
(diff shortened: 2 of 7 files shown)


LucasArmandVast:
This seems mostly fine to me. The versioning code using the git describe subprocess seems a bit strange to me... I guess it would be needed if importlib doesn't work when running from a git branch, so probably fine. Other than that, I think you should smoke-test all paths (SDK, CLI, serverless) with the version on this branch

madisonvast:
@LucasArmandVast 
Good catch on the git-describe code — it's actually pre-existing, relocated from cli/util.py to break a circular import, but your instinct was right: it ran git describe in the process CWD, so it could pick up tags from an unrelated repo. Fixed by pinning cwd to the package dir. Smoke-tested all three paths (SDK, CLI, serverless) from this branch against a local webserver — all send the right UA and get labeled SDK/CLI/SDK respectively. Note the version reads 0.0.0 in dev poetry installs since dynamic versioning only stamps at build time; published installs report the real version.

LucasArmandVast (approved):

PR #420, opened 2026-06-12
feat(CLN-3413): read GPU names + VRAM from /api/v0/gpu_types/ with hardcoded fallback

Marcoh572:
## What
Migrates the `vastai` CLI package off its hardcoded GPU data to the `/api/v0/gpu_types/` catalog, with the existing hardcoded data retained as a backwards-compatible fallback.

- New `_get_gpu_types()` in `cli/util.py` — 24h-cached anonymous fetch, mirrors `_get_gpu_names()`, returns `None` on error.
- `benchmarks.py` resolves canonical GPU names from the **union** of the `query.py` constants and the catalog, and reads per-card VRAM from the catalog's `gpu_ram_mb` with fallback to `_DEFAULT_GPUS`.
- The 4-GPU default sweep set is unchanged (editorial choice, no catalog equivalent).

## Why
Part of the CLN-3375 epic — make the `gpu_types` table the single source of truth and have everything read the public endpoint. Adding a SKU via the admin API then surfaces on the CLI within the cache cycle, with no code change.

## Backwards compatibility
Nothing is deleted. When the catalog is unreachable or a field is null, behavior is identical to today (constants + `_DEFAULT_GPUS`). `gpu_ram_mb` is currently unpopulated, so VRAM falls back to the dict now and auto-upgrades when CLN-3449 backfills it.

## Scope
`vastai/` package only. `vast.py` is untouched — it already fetches GPU names dynamically and carries no constants/VRAM dict.

## Test plan
- New `tests/cli/test_benchmarks_gpu_catalog.py`: union adds SKUs, fallback when catalog unreachable, `_parse_gpu_spec` resolves a catalog-only name, VRAM override / null-fallback / unreachable-fallback / unknown→None.
- Autouse guard in `test_benchmarks_commands.py` keeps the existing suite hermetic (no live HTTP from the new fetch).
- Note: suite not run in the authoring env (no Poetry install there); logic validated via a standalone harness and byte-compile. Please confirm `poetry run pytest tests/cli/` is green in CI.

Jira: CLN-3413

### DIFF
--- a/tests/cli/test_benchmarks_commands.py
+++ b/tests/cli/test_benchmarks_commands.py
@@ -14,6 +14,19 @@ import pytest
 from vastai.cli.commands import benchmarks as bench
 
 
+@pytest.fixture(autouse=True)
+def _stub_gpu_catalog():
+    """Force the catalog fetch in _parse_gpu_spec to fall back to the hardcoded
+    constants so no live HTTP call leaks into these tests.
+    """
+    bench._canonical_gpu_names.cache_clear()
+    bench._catalog_vram_map.cache_clear()
+    with patch.object(bench, "_get_gpu_types", return_value=None):
+        yield
+    bench._canonical_gpu_names.cache_clear()
+    bench._catalog_vram_map.cache_clear()
+
+
 # ---------------------------------------------------------------------------
 # _benchmark_gpu — cleanup invariant
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


abiola-vastai (approved):

vastzuby (approved):

vastzuby (approved):
looks good to me. though maybe you should wait to merge this until the gpu types endpoint is deployed

PR #425, opened 2026-06-16
CLN-3459: make vast.py a thin shim delegating to vastai.cli.main

sammy-vastai:
**Problem:** The repo shipped two parallel CLI implementations with no shared code — the ~9.8k-line `vast.py` monolith (`python vast.py …`) and the `vastai` package (`vastai …`, the pip entry point). They drift independently, so a fix in one isn't in the other, and it's unclear which is the source of truth.

**Change:** `vast.py` is now a ~22-line compatibility shim that delegates to `vastai.cli.main` — the same import the `vastai` entry point uses. The `vastai` package is the single source of truth; `python vast.py` keeps working unchanged.

**Why it's safe:**
- No live commands were lost. The three commands the diff flagged as "old-only" (`create account`, `generate pdf-invoices`, `vm copy`) are all already dead code in `vast.py` (commented out or wrapped in string literals).
- Nothing in scripts/tests/docs imports *from* `vast.py`.
- Preserves the `# PYTHON_ARGCOMPLETE_OK` marker (tab-completion) and `KeyboardInterrupt`/`BrokenPipeError` handling. Adds a `sys.path.insert` so it works from a bare checkout with no `pip install`.

**Verification:**
- `python vast.py --version` / `--help` / subcommand help all work and match the package.
- Test suite identical with and without the change (105 pre-existing env failures, 1070 passed — none introduced).

Design notes: `docs/single-source-of-truth.md`.

**Follow-up (not in this PR):** update README / `vastai/SKILL.md` to document `vastai` as canonical and `python vast.py` as a supported shim.


sammy-vastai:
Some tooling (daemon update scripts, verifications scripts) still relies on fetching just the vast.py file, so this would break that.  Deferring until bash installer is in

PR #427, opened 2026-06-19
feat(installer): musl support + glibc floor + OS-matrix CI

sammy-vastai:
Closes the OS-matrix gap in the installer rollout (install-design.md §12 step 5). Building the matrix immediately surfaced two real `install.sh` bugs, both fixed here, plus a dependency change needed to actually support musl.

## What broke (found by running install.sh across real OS containers)

| Target | Before | After |
|---|---|---|
| debian-slim (no Python) | works | works |
| ubuntu 20.04+/debian 11+ (glibc >=2.31) | works | works |
| **alpine / musl** | glibc uv installed -> `cannot execute` | installs |
| **old glibc (<2.31)** | cryptic mid-build dep failure | clean bail to pip |

## Fixes

**1. musl detection (`set -o pipefail` bug).** `ldd --version | grep musl` returned non-zero on Alpine because musl's `ldd` exits 1 on `--version`; with `pipefail` that discarded a *successful* grep, so musl went undetected and the glibc uv binary was installed. Now keys off the `ld-musl-*` loader directly (exit-code-free), with a `|| true`-guarded `ldd` fallback.

**2. glibc floor.** Below glibc 2.31 (Ubuntu 20.04+/Debian 11+), deps lack wheels for the pinned CPython and would fail mid-build. `install.sh` now detects glibc and exits cleanly pointing at pip *before* touching `\$ROOT`. Overridable via `VASTAI_GLIBC_FLOOR` (also a hermetic-test seam).

**3. `psutil ~=6.0` -> `~=7.0`.** psutil <7 ships **no musllinux wheels**, so Alpine could never install wheels-only. 7.x has them. Verified the *entire* dep closure (pycares, cryptography, pillow, aiohttp, tokenizers) resolves binary-only on musl. Only codebase usage is `psutil.disk_usage()` - unchanged across 6->7. `poetry.lock` re-locked (psutil 6.1.1->7.2.2, no other bumps).

## Tests / CI
- New hermetic test `glibc_floor` (asserts clean bail + zero residue); all 6 pass locally, shellcheck clean.
- **`installer-ci.yml`**: new `build-wheel` + `os-matrix` jobs - install *this PR's* wheel via the real `install.sh` across `alpine:latest`, `debian:stable-slim`, `ubuntu:20.04`, and `macos-latest`. Container legs run via `docker run` on the ubuntu host (checkout stays on host -> avoids the actions/checkout-on-musl node problem); macOS runs natively. Triggers now include `pyproject.toml` so dep changes (the psutil class of bug) are gated. Non-blocking (`continue-on-error`) until proven green, then flip to gating.
- **`python-publish.yml`**: added `macos-latest` to the post-publish smoke matrix (real artifacts).

## Validation
Every scenario above was validated locally in Docker against the live `install.sh` before writing CI. macOS legs are validated only in CI (no local Darwin).


vastzuby (approved):
looks good. claude flagged one thing, though nonblocking 

```
. (Medium) The glibc-floor CI leg tests the pass case, not the bail. .github/workflows/installer-ci.yml — the matrix entry { name: 
  glibc-floor, image: "ubuntu:20.04" } ships glibc 2.31, which equals GLIBC_FLOOR=2.31, and equal-to-floor passes ([ "$v" != "$GLIBC_FLOOR" ]
  is false, no die). So this leg only verifies a normal install on the minimum supported glibc; the actual bail-to-pip path is exercised
  solely by the hermetic unit test (which fakes it with VASTAI_GLIBC_FLOOR=99.0). If someone inverted the sort -V comparison, CI would stay
  green because no matrix image is below 2.31. Consider a sub-floor image (e.g. ubuntu:18.04) asserting the clean pip-pointing exit, or rename
  the leg to reflect that it tests floor-equal.
```

PR #431, opened 2026-06-23
feat(deployments): support workload_calculator for remote functions

vastzuby:
adds an optional argworkload calculator arg to remote functions for deployments. 

the workload calculator tells us the work per request. for deployments, we weren't using this calculator and defaulting to 100. so autoscaler was scoring all remote functions the same regardless of actual cost

I also added `wrap_workload_calculator` so the user can set the workload calculator as python but the request reaches the worker as json. a similar function `wrap_remote_func` already exists but we cant use it because its async and returns serialized results (we want deserialized)

### DIFF
--- a/vastai/serverless/remote/deploy.py
+++ b/vastai/serverless/remote/deploy.py
@@ -309,6 +309,7 @@ class Deployment(Deployment_):  # TODO: Async Context Manager compatible with cl
         benchmark_dataset: list[dict] | None = None,
         benchmark_generator: Callable[[], dict] | None = None,
         benchmark_runs: int = 10,
+        workload_calculator: Callable[..., float] | None = None,
     ) -> (
         Callable[P, Awaitable[Any]]
         | Callable[[Callable[P, Awaitable[Any]]], Callable[P, Awaitable[Any]]]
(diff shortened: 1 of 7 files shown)


vastai/serverless/remote/deploy.py
```diff
@@ -309,10 +309,14 @@ def remote(
...
         benchmark_generator: Callable[[], dict] | None = None,
         benchmark_runs: int = 10,
+        workload_calculator: Callable[..., float] | None = None,
     ) -> (
         Callable[P, Awaitable[Any]]
         | Callable[[Callable[P, Awaitable[Any]]], Callable[P, Awaitable[Any]]]
     ):
+        # workload_calculator is consumed on the worker (serve mode); on the
+        # client/deploy side the decorator only sets up remote dispatch, so it
+        # is accepted here purely for signature parity.
```
LucasArmandVast: remove comment

vastzuby: :+1:

vastzuby:
@LucasArmandVast I tested this change with two of the deployment examples. also added workload_calculator to them in `/examples/deployments`

LucasArmandVast (approved):

PR #424, opened 2026-06-16
fix version3

stackanthony:

PR #345, opened 2026-03-25
Consolidate SDK and CLI into unified vastai package

vastzuby:
replaces the monolithic vast.py with a modular vastai/ package. 
one pip package for both the cli and sdk. 
everything is backwards and forwards compatible.

splits vast.py into its own modules:
- cli commands
- api layer
- sdk
- async/sync clients
- data models

also adds:
- vastai_sdk/ for a backward-compat shim so existing imports still work
- sdk-wrapper/ for publishing vastai-sdk to pypi
- full test suite (401 tests), workflow will run on PRs
- new deployments feature, 2fa stuff
- two-job ci workflow (publishes vastai + vastai-sdk to pypi)

`vast.py` is still there as a deprecated shim for backwards compatibility

PR #434, opened 2026-06-23
chore(deps): purge unused transformers dependency (CLI + tests)

sammy-vastai:
## Why

A CLI should not carry an ML framework. `transformers` is imported **nowhere** in `vastai/`, `vastai_sdk/`, or `vast.py` (the only repo references are in `examples/deployments/*`, which `pip_install` it into *remote* deployment images, not the CLI). It was a hard dependency for no functional reason — and it anchors a large transitive cluster (`numpy`, `tokenizers`, `huggingface-hub`, `hf-xet`, `safetensors`, `regex`) that nothing else needs.

### History (reconciling the Dependabot context)
`transformers` reached us two ways, both stale:
- **`tests/poetry.lock`** carried `transformers~4.52` (vulnerable, `<4.53.0`) via an old fat `vastai-sdk` footprint — the source of **5 open Dependabot ReDoS alerts**. The `transformers>=4.53.3` pin added in #424 was the floor-bump remediation.
- The root `vastai` pin then propagated the dep forward even though nothing imported it.

The tests lock had been **frozen stale since #345** (SDK/CLI consolidation): `tests/pyproject.toml` declared `vastai-sdk = { path = ".." }`, but `..` is the **`vastai`** package — so `poetry lock` errored on the name mismatch and could never regenerate (CI uses the committed lock without re-resolving, hiding it).

## Changes
1. **Root** — remove `transformers>=4.53.3` from `pyproject.toml`; `poetry lock` drops the entire cluster incl. `numpy` (766 lines gone from `poetry.lock`).
2. **Tests** — fix the path-dep name (`vastai-sdk` → `vastai`, matching the package at `..` and the `from vastai` imports) so the lock can regenerate; re-lock purges `transformers`, `numpy`, `tokenizers`, plus unused `nltk`/`fastapi`/`uvicorn`/`hf-transfer` cruft.

## Impact
- **~138M (~47%)** off the managed install venv (transformers 51M + numpy 57M + hf_xet 12M + tokenizers 10M + …).
- **Closes the 5 `transformers` Dependabot alerts** (all on `tests/poetry.lock`).
- Fixes the long-broken `tests/` re-lock (name mismatch since #345).

## Validation
- `pytest -m 'not live and not integration'`: **1178 passed**. The only failures (12 in `test_tar_utils.py`) are a **pre-existing** `compress=` kwarg drift on `master` — unrelated to this change (a removed dep would surface as `ImportError`, not an unexpected-kwarg `TypeError`; this branch doesn't touch that code).
- `pyproject.toml` changes trigger the `installer-ci` OS matrix (alpine/musl + macOS) to confirm cross-platform install.


stackanthony (approved):

sammy-vastai:
@stackanthony The root cause was caused by tests/poetry.lock not being regenerated after we found that this dependency was not necessary (@vastzuby https://github.com/vast-ai/vast-cli/pull/345#issuecomment-4202896693)

PR #442, opened 2026-07-01
installer: Rosetta 2 detection

sammy-vastai:
Small hardening item for the installer, borrowed from a comparison with the Claude Code installer.

## Rosetta 2 detection (`scripts/install.sh`)

An ARM Mac running an x86_64 shell (Rosetta) previously got the x86_64 uv/CPython — works, but emulated. `detect_platform` now checks `sysctl -n sysctl.proc_translated` on Darwin x86_64 and switches to `DARWIN_ARM64` for the native build. The sysctl key is absent on Intel Macs, so the probe is empty there and it's a no-op; guarded with `|| true` per the script's `set -e` conventions.

## Testing

- `tests/installer/run_tests.sh` — 8 passed, 0 failed (Rosetta branch is Darwin-only; Linux behavior unchanged)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -133,6 +133,12 @@ detect_platform() {
         aarch64|arm64) [ "$os" = "DARWIN" ] && arch="ARM64" || arch="AARCH64" ;;
         *) die "unsupported architecture: $arch — install with pip instead: pip install vastai" ;;
     esac
+    # An x86_64 shell under Rosetta 2 reports the emulated arch; install the
+    # native arm64 build instead. (sysctl key absent on Intel Macs → empty → no-op.)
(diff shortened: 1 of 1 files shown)


vastzuby (approved):
howd you test this

sammy-vastai:
> howd you test this

@vastzuby So I haven't fully tested this E2E.  But this same check is in the claude code installer so should be safe.  The one thing blocking release of this is just real OS testing (intel macs, arm macs, different linux distros).  I've tested all the linux distros through docker containers, so the only thing remaining is mac testing.   Will do a smoke test on my personal machine, and if that works, I think we should be in good shape for the initial/quiet release of this bash installer.

PR #443, opened 2026-07-02
fix(installer): guard empty array expansion for macOS bash 3.2

sammy-vastai:
## Problem

Interactive installs on macOS fail:

```
$ curl -fsSL https://vast.ai/install.sh | bash
vastai-install: Installing vastai 1.1.3 (Python 3.12, isolated in /Users/.../.vastai)...
bash: line 192: quiet[@]: unbound variable
```

`install_version` empties the `pyquiet` array when stderr is a TTY (to show uv's Python download progress interactively), then expands it with `"${pyquiet[@]}"`. On bash < 4.4, expanding an empty array under `set -u` is an "unbound variable" error — and macOS pipes `curl | bash` into stock `/bin/bash` 3.2.

The bug needs both **bash 3.2** and **a TTY on stderr**, which is why the `macos-latest` job in installer CI never caught it: CI has no TTY, so the array stays non-empty and the failing path never runs.

## Fix

Use the bash-3.2-safe guard idiom:

```bash
${pyquiet[@]+"${pyquiet[@]}"}
```

which expands to nothing when the array is empty instead of tripping `set -u`. Verified under `bash -u`: empty → no extra args, non-empty → passes `--quiet`. The other arrays in the script (`vflags`, `wflags`) are always non-empty, so this was the only unsafe expansion.

Note: users hit this via the release-attached bootstrap, so the live installer stays broken for interactive macOS installs until the next release re-uploads `install.sh`.

### DIFF
--- a/.github/workflows/installer-ci.yml
+++ b/.github/workflows/installer-ci.yml
@@ -27,8 +27,16 @@ jobs:
       - name: python syntax check
         run: python3 -m py_compile scripts/make_manifest.py scripts/publish_release.py
 
+  # macOS matters here beyond os coverage: its stock /bin/bash is 3.2, which
+  # is what test_tty_install needs to catch set -u empty-array regressions
+  # (fatal on bash < 4.4, invisible on Linux CI's modern bash).
   hermetic-tests:
-    runs-on: ubuntu-latest
+    name: hermetic-tests (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - name: run install.sh hermetic scenarios (no external network)
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -202,9 +202,11 @@ install_version() {
     # uv venv's own "Activate with: source .../.env.new/..." line is misleading
     # here — .env.new is renamed to env/ below, and vastai is a symlinked CLI you
     # run, never a venv you "activate".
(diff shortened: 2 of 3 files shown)


sammy-vastai:
Added a regression test that fails before the fix and passes after it.

**`test_tty_install`** (hermetic suite): runs the real installer on a pty via `script(1)`, so `[ -t 2 ]` is true and `install_version` takes the branch that empties `pyquiet` — the exact condition of the field failure. The bug is only fatal on bash < 4.4, so the test runs the installer under `${VASTAI_TEST_TTY_BASH:-/bin/bash}`; the regression guarantee comes from hosts whose `/bin/bash` predates 4.4. To that end, the hermetic-tests CI job now also runs on `macos-latest`, whose stock `/bin/bash` is 3.2.

Verified locally against a bash 3.2.57 built from source:

```
=== PRE-FIX (master's install.sh) ===
    assert failed: no set -u explosion at a TTY
FAIL tty_install  (output below)
    |   Installing vastai 1.2.3 (Python 3.12) → .../sb-tty/root
    | scripts/install.sh: line 207: pyquiet[@]: unbound variable
0 passed, 1 failed

=== POST-FIX (this branch's install.sh) ===
PASS tty_install
1 passed, 0 failed
```

Also made the harness itself bash-3.2-clean, since the macOS leg runs `run_tests.sh` under 3.2: replaced `SELECTED=("${@:-…}")` (expanding `$@` with zero args under `set -u` is also fatal on old bash) and reaped the fixture server in `cleanup` (bash 3.2 prints a stray "Terminated" otherwise). Full 9-scenario suite passes under both bash 5.2 and bash 3.2.57, and shellcheck `-S style` is clean.

vastzuby (approved):

PR #455, opened 2026-07-13
fix(deps): remediate flagged CVEs in cryptography, urllib3, lxml

stackanthony:
## Summary

Remediates the GitHub Dependabot vulnerabilities flagged against **vast-cli** by bumping the affected packages to their fixed versions in both dependency manifests and both lock files.

| Package | Was | Now | Fixes |
|---|---|---|---|
| `cryptography` | 46.0.5 | **49.0.0** | GHSA-537c-gmf6-5ccf (`< 48.0.1`) |
| `urllib3` | 2.6.3 | **2.7.0** | CVE-2026-44432 & CVE-2026-44431 (`< 2.7.0`) |
| `lxml` | 5.3.2 / 6.0.2 | **6.1.1** | CVE-2026-41066 (`< 6.1.0`) |

## Changes

- `pyproject.toml` — bumped direct deps: `urllib3>=2.0,<3.0` → `>=2.7.0,<3.0`, `cryptography==46.0.5` → `==49.0.0`.
- `tests/pyproject.toml` — raised `cryptography` floor `>=41.0` → `>=48.0.1` (above the vulnerable range).
- `poetry.lock` + `tests/poetry.lock` — re-locked. `lxml` is transitive (via `borb`, which permits `>=4.9.1`), so it's fixed purely in the lock.

Actual version changes are scoped to the three targets plus a required `pyopenssl` 26.0.0 → 26.3.0 bump for cryptography-49 compat in the tests lock. The large line count in the lock diffs is just poetry 2.x reformatting/rehashing.

## Verification

- `poetry check --lock` passes on both projects.
- Updated deps install cleanly; `cryptography 49.0.0`, `lxml 6.1.1`, `urllib3 2.7.0` all import, and `vastai`/`vastai_sdk` import fine.
- Unit suite: **1190 passed** with the new versions (including the `conftest.py` x509/RSA cryptography paths).
- The 12 pre-existing `test_tar_utils.py` failures (`create_deployment_tarball()` `compress` kwarg mismatch) are unrelated — this PR touches no source or test files.

> Note: `pillow` / `aiohttp` were flagged only for the separate **vast-sdk** / **base-image** products; in this repo they're already at safe versions (`pillow==12.2.0`, `aiohttp>=3.13.3`), which is why they weren't in the vast-cli alerts.


LucasArmandVast (commented):
```
Yes, there are several. You're crossing three major releases (47, 48, 49), and pyca ships BACKWARDS INCOMPATIBLE changes in each. Here's what actually breaks going 46.0.5 → 49.0.0:

**Environment/platform:**
- Python 3.8 support removed in 48.0.0 — Python 3.9+ now required
- OpenSSL 1.1.x support removed in 47.0.0 — OpenSSL 3.0.0+ required (only matters if you build from source against system OpenSSL; wheels bundle their own)
- x86_64 macOS wheels removed in 49.0.0 (arm64 only) and 32-bit Windows dropped — irrelevant for Linux servers, but could bite Intel Mac dev machines or CI runners
- 47.0.0 bumped MSRV to Rust 1.83 (again, source builds only)

**API removals/behavior changes:**
- Binary elliptic curves (`SECT*` classes) removed in 47.0.0 — rarely used, probably not you
- Deprecated type aliases removed in 49.0.0: `PUBLIC_KEY_TYPES`, `PRIVATE_KEY_TYPES`, `CERTIFICATE_PRIVATE_KEY_TYPES`, etc. — use `PublicKeyTypes`, `PrivateKeyTypes`, and friends instead. Worth grepping for; these show up in type annotations in older code.
- Loading keys with unsupported algorithms now raises `UnsupportedAlgorithm` instead of `ValueError` in `load_pem_private_key()` and related functions — check any `except ValueError` around key loading
- `public_bytes`/`private_bytes` now raise `TypeError` instead of `ValueError` for invalid encoding/format combos
- ChaCha20 (raw, non-AEAD) now treats the first 4 bytes of the nonce as an RFC 7539 block counter and raises `ValueError` on counter overflow — only affects direct `ChaCha20` cipher use, not `ChaCha20Poly1305`
- Stricter X.509 parsing: CRLs with mismatched inner/outer signature algorithms now fail at load time (48.0.0), and certs with encoded NULL params in ECDSA/DSA AlgorithmIdentifiers (old Java-emitted certs) now raise `ValueError` instead of loading with a warning

**Deprecations to watch (not yet removed):**
- CFB/OFB/CFB8 modes and Camellia moved to the `decrepit` module in 47.0.0, slated for removal from the standard modules — these were still importable from the normal locations, so you'll get warnings, then breakage at 50
- TripleDES with 64-bit or 128-bit keys deprecated; only 192-bit keys will be accepted in a future release

Also check transitive constraints: the latest pyOpenSSL requires cryptography >= 49.0.0, so if anything in your dep tree pins an older pyOpenSSL, pip's resolver may fight you. Same story with paramiko/boto3-adjacent stuff occasionally.

For a typical Pyramid webserver doing TLS, key loading, and maybe Fernet/JWT, the realistic risks are the type alias removals, exception-type changes around key loading, and pyOpenSSL version coupling. A quick `grep -rE 'PUBLIC_KEY_TYPES|PRIVATE_KEY_TYPES|SECT|Camellia|modes\.(CFB|OFB)'` plus running the test suite with `-W error::CryptographyDeprecationWarning` on 46.0.5 first will surface most of it before you bump.
```

Should be tested before merging

stackanthony:
@LucasArmandVast The cryptography module is only ever used for tests. If those pass, we're good

LucasArmandVast (approved):

PR #462, opened 2026-07-20
Client/host command classification with role-aware CLI display

sammy-vastai:
## Summary
- Classifies every CLI command as host-only or not (`is_host_only_command()` in `vastai/cli/parser.py`), replacing the 26 hand-written `[Host]` help-text prefixes with an auto-rendered one derived from the classification.
- Adds a stored client/host role (`~/.config/vastai/vast_role`) that filters `--help`, tab completion, and error hints — display-only, never gates execution; the server remains the real permission enforcement layer.
- **Client is the default.** Host-only commands only show once the role is explicitly resolved to `host` (`vastai set role host`, or auto-detection in the follow-up PR). An unset or unrecognized role is treated as client, not "show everything" — this is what actually makes fresh installs show the filtered view without requiring the auto-detection in #463.
- `vastai set role host|client` to manually set/override the role.
- A client-role user hitting a `machine_read`-style 401 on a host command gets a "did you mean" hint (e.g. `show machines` → `show instances`).

Auto-detecting the role from an existing API key (rather than requiring `set role` to be run manually) is split into a follow-up stacked PR: #463.

## Test plan
- [x] `poetry run pytest tests/cli` — 397 passed
- [x] Manually verified: a completely fresh install (no role file, no key) shows the filtered client view by default; `set role host`/`set role client` toggle correctly; the 401 hint fires even with an unset role
- [x] Confirmed pre-existing failures in `test_tar_utils.py`/`test_search.py` are unrelated (present on `master` too)

CLN-3582

### DIFF
--- a/tests/cli/test_auth_commands.py
+++ b/tests/cli/test_auth_commands.py
@@ -119,6 +119,28 @@ class TestSetApiKey:
             assert MockClient.call_args[0][0] == "round-trip-key"
 
 
+class TestSetRole:
+    def test_set_role_host(self, tmp_path, monkeypatch, capsys):
+        role_file = tmp_path / "vast_role"
+        monkeypatch.setattr("vastai.cli.util.ROLE_FILE", str(role_file))
+        from vastai.cli.commands.auth import set__role
+        set__role(argparse.Namespace(role="host"))
+        assert role_file.read_text() == "host"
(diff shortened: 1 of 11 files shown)


sammy-vastai:
Superseded by #484, which combines this and #463 into a single branch (rebased onto current master, conflicts resolved, tests passing).

PR #463, opened 2026-07-20
Auto-detect the client/host role from an existing API key

sammy-vastai:
## Summary

Stacked on #462. Adds automatic role detection so most users never have to run `vastai set role` manually — the key insight is that for most existing users an API key already exists, so there's no "enter your key" moment to hook into; detection just needs to use it as soon as possible.

- `ensure_host_role_detected()` (`vastai/cli/util.py`): one best-effort API call (`GET /machines?owner=me`) that caches the role permanently either way ('client' as much as 'host') on success; a failed check (network/auth error) leaves it undetected for retry rather than guessing.
- Wired into three trigger points that already have a valid key in hand:
  - `get_client()` (`vastai/cli/utils.py`) — fires lazily on a user's next real command. Deliberately does not run on `--help`/tab completion, which never build a client.
  - `vastai set api-key` — detects immediately using the newly-saved key and announces the result.
  - The bash installer (`scripts/install.sh`) — if a key already exists on disk (a pre-existing user re-running the installer to upgrade), fires a single detached (`nohup ... & disown`) background call so it can't add install latency.
- pip installs still can't be reached at install time (no post-install hook exists in the wheel/PEP 517 standard) — they fall back to the `get_client()` lazy path, an accepted, unavoidable gap.
- Once resolved, the role is sticky — `vastai set role host|client` remains the deliberate override for e.g. a client account that starts hosting later.

## Test plan
- [x] `poetry run pytest tests/cli` — 407 passed
- [x] `bash tests/installer/run_tests.sh` — 18 passed (3 new scenarios covering the installer's background role-detection)
- [x] Manually verified `set api-key` auto-detection and announcement against the real `vastai` binary

CLN-3582

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -328,6 +328,23 @@ check_pip_coexistence() {
     warn "put $LOCAL_BIN first in PATH so this one wins."
 }
 
+# warm_role_cache — best-effort: most reinstalls are an existing user with an
+# API key already on disk from before, so there's no "enter your key" moment
+# to hook into here — just a key that's already sitting there. If one exists,
+# resolve the client/host CLI view for it now (one GET to /machines) instead
+# of waiting on whatever command they happen to run next. A fresh install has
+# no key yet, so this is a no-op for that (the common) case.
+#
+# Detached (nohup, backgrounded, disowned) so it can't add latency to the
+# install and survives this script's own process exiting; every failure mode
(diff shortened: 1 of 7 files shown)


sammy-vastai:
Superseded by #484, which combines this and #462 into a single branch (rebased onto current master, conflicts resolved, tests passing).

PR #464, opened 2026-07-21
Un-deprecate show instances: absorb API version churn instead of surfacing it

sammy-vastai:
## Summary

Fixes [CLN-3581](https://vastai.atlassian.net/browse/CLN-3581): `vastai show instances` was printing a deprecation warning pointing users at `show instances-v1`, even though the v0→v1 backend migration was already done transparently at the transport layer. Command names are the CLI's stable contract, so this folds the v1 improvements into the existing command instead of shipping a versioned replacement.

- Removed the `DEPRECATED` warning; `show instances` now uses `/api/v1/instances/` directly.
- Folded `instances-v1`'s server-side filtering (`--status`, `--label`, `--gpu-name`, `--verification`), sorting (`--order-by`), custom columns (`--cols`), and rich table output into `show instances`.
- `show instances-v1` is now a hidden alias of `show instances` (no longer listed in `--help`), so existing scripts keep working.
- `VastAI.show_instances()` no longer emits a `DeprecationWarning`.
- Kept the merge backward compatible by contract:
  - `--raw` still returns a flat JSON list of *all* instances by default (fetches every page silently); it only returns the paginated `{instances, next_token, total_instances, label_counts}` dict when the caller explicitly passes `--limit`/`--next-token`.
  - No pagination prompts by default — fetch-all is silent. The interactive "Fetch next page?" prompt only appears in explicit single-page mode, and only when stdin/stdout are a TTY, so non-interactive invocations (cron, agents) never block.
- Generalized `instances_api.show_instances()` to accept optional `select_filters`/`order_by` so the CLI can reuse it for the filtered fetch-all `--raw` path.
- Updated `vastai/SKILL.md` examples to drop `instances-v1` in favor of the merged command.

## Test plan

- [x] `poetry run pytest tests/cli/test_instances_commands.py tests/sdk/test_sync_client.py tests/cli/test_parser.py` — 60 passed
- [x] `poetry run pytest tests/ --ignore=tests/live` — no new failures (pre-existing unrelated failures in `test_tar_utils.py`/`test_search.py` confirmed present on `master` too)
- [x] Manual verification against the live API via `poetry run vastai`:
  - `show instances` — no deprecation warning, rich table with filter/summary panels
  - `show instances --raw` — flat list, legacy shape
  - `show instances -q` — bare IDs
  - `show instances --status running --gpu-name 'RTX 4090'` — filters work
  - `show instances-v1 -q` — hidden alias still functions
  - `vastai --help` — `instances-v1` no longer listed
  - Simulated non-TTY explicit-pagination mode — confirmed no blocking prompt

### DIFF
--- a/tests/cli/test_instances_commands.py
+++ b/tests/cli/test_instances_commands.py
@@ -67,6 +67,164 @@ class TestShowInstances:
         result = args.func(args)
         assert isinstance(result, list)
 
+    def test_show_instances_no_deprecation_warning_stderr(self, parse_argv, patch_get_client, mock_response, capsys):
+        patch_get_client.get.return_value = mock_response(200, {"instances": [], "next_token": None})
+        args = parse_argv(["show", "instances", "--raw"])
+        args.func(args)
+        captured = capsys.readouterr()
+        assert "DEPRECATED" not in captured.err
+
+
+class TestShowInstancesFilters:
+    def test_status_filter_sent(self, parse_argv, patch_get_client, mock_response):
+        patch_get_client.get.return_value = mock_response(200, {"instances": [], "next_token": None})
+        args = parse_argv(["show", "instances", "--raw", "--status", "running", "loading"])
+        args.func(args)
+        select_filters = patch_get_client.get.call_args.kwargs["query_args"]["select_filters"]
+        assert select_filters == {"actual_status": {"in": ["running", "loading"]}}
+
+    def test_invalid_status_warns(self, parse_argv, patch_get_client, mock_response, capsys):
+        patch_get_client.get.return_value = mock_response(200, {"instances": [], "next_token": None})
+        args = parse_argv(["show", "instances", "--raw", "--status", "bogus"])
+        args.func(args)
+        captured = capsys.readouterr()
+        assert "unknown status value" in captured.err
+
+    def test_label_filter_empty_string_means_unlabeled(self, parse_argv, patch_get_client, mock_response):
+        patch_get_client.get.return_value = mock_response(200, {"instances": [], "next_token": None})
+        args = parse_argv(["show", "instances", "--raw", "--label", ""])
+        args.func(args)
+        select_filters = patch_get_client.get.call_args.kwargs["query_args"]["select_filters"]
+        assert select_filters == {"label": {"in": [None]}}
+
+    def test_gpu_name_filter_sent(self, parse_argv, patch_get_client, mock_response):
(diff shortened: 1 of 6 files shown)


sammy-vastai:
@Marcoh572 Could you take a look at this PR when you get the chance?

Marcoh572 (approved):
This looks great — folding v1 back into `show instances` and hiding the old name reads cleanly, and the fetch-all-by-default plus the TTY-gated prompt all line up with what the ticket asked for. Approving. Two things, neither blocking:

1. The acceptance criteria don't really have tests behind them yet — the suite covers `--raw` and the basic table, but nothing asserts the deprecation warning is gone, that `-q` prints bare IDs, that `instances-v1` still routes as a hidden alias, or that a non-TTY run never blocks on the pagination prompt. Might be worth adding a few so they don't quietly regress later — especially the no-warning and no-prompt ones, since those are the whole point of the change.

2. A behavior question: now that fetch-all is the default, a bare `show instances` runs through the main loop with the 1s sleep between pages, so someone with a lot of instances (~40 pages) waits ~40s — where the old default didn't throttle at all. `--raw` skips it since it goes through the api-layer pager. Was the sleep meant to sit on the default human-facing path? Could it move to the `--raw`-style fast path or drop to a shorter interval? Probably doesn't matter for typical accounts, just flagging it.

sammy-vastai:
> This looks great — folding v1 back into `show instances` and hiding the old name reads cleanly, and the fetch-all-by-default plus the TTY-gated prompt all line up with what the ticket asked for. Approving. Two things, neither blocking:
> 
> 1. The acceptance criteria don't really have tests behind them yet — the suite covers `--raw` and the basic table, but nothing asserts the deprecation warning is gone, that `-q` prints bare IDs, that `instances-v1` still routes as a hidden alias, or that a non-TTY run never blocks on the pagination prompt. Might be worth adding a few so they don't quietly regress later — especially the no-warning and no-prompt ones, since those are the whole point of the change.
> 2. A behavior question: now that fetch-all is the default, a bare `show instances` runs through the main loop with the 1s sleep between pages, so someone with a lot of instances (~40 pages) waits ~40s — where the old default didn't throttle at all. `--raw` skips it since it goes through the api-layer pager. Was the sleep meant to sit on the default human-facing path? Could it move to the `--raw`-style fast path or drop to a shorter interval? Probably doesn't matter for typical accounts, just flagging it.

thanks for the review.  added some tests.  did not intend to make sleeping between requests the default, couldn't find a reason for it so just dropped it for consistency.

PR #466, opened 2026-07-21
Hide unreleased network-volume commands from --help

sammy-vastai:
## Summary

Network volumes aren't released yet, but their CLI commands were fully visible in `vastai --help` and tab completion — `search network-volumes`, `create network-volume`, `list network-volume`, `unlist network-volume`.

Adds a general-purpose "hidden command" gate, independent of and unrelated to the client/host work in #462/#463:
- `HIDDEN_COMMANDS` (name-based registry) + an explicit `hidden=` kwarg on `@parser.command`, mirroring the existing `COMMAND_OVERRIDES`-style pattern in `vastai/cli/parser.py`.
- Discoverability gate only, not an access gate — a hidden command still runs if typed directly (e.g. `vastai search network-volumes ...`), so internal testing is unaffected.
- Applied to all four network-volume commands.

Found while manually testing #462 — noticed `search network-volumes` in `--help` but its own `usage:` string has an unrelated pre-existing typo (says `search network volumes`, missing the hyphen, present on `master` too). Not fixed here since it's a separate, small, unrelated issue — happy to do it as a follow-up if wanted.

## Test plan
- [x] `poetry run pytest tests/cli` — 375 passed
- [x] Manually verified: none of the four commands appear in `vastai --help` or tab completion; all four still parse and run when typed directly

### DIFF
--- a/tests/cli/test_parser.py
+++ b/tests/cli/test_parser.py
@@ -4,6 +4,7 @@ import pytest
 from vastai.cli.parser import (
     apwrap, argument, hidden_aliases, MyWideHelpFormatter,
     build_command_maps, two_stage_command_completions,
+    is_hidden_command,
 )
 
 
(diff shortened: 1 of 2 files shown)


vastzuby (approved):

vastzuby:
yeah anything network volume or cluster related should be hidden

PR #468, opened 2026-07-22
feat(serverless): health-gated readiness mode (readiness='healthcheck')

robballantyne:
## What

Adds an **opt-in** readiness mode to the PyWorker `Backend` so a worker can mark the model loaded on the first successful `/health` probe, instead of keying off a backend-specific log line (vLLM's `"Application startup complete."`). This makes the `openai` worker generic across vLLM / SGLang / llama.cpp.

Part of the serverless-enablement work (CON-1611, WS2). **Draft** — companion changes land separately (see Follow-ups).

## Changes

- `WorkerConfig` + `Backend` gain:
  - `readiness`: `'logs'` (default) | `'healthcheck'` — validated in `__post_init__` (a typo raises, can't silently fall through).
  - `readiness_timeout` (300.0) — the wait-for-first-200 budget; also drives the existing log/lifecycle health waits (same default, no behavior change).
  - `healthcheck_probe_timeout` (10.0) — per-probe `/health` HTTP timeout, replacing a hardcoded `ClientTimeout(total=10)`. Set `>=` the backend's own health timeout (SGLang's `/health` can take ~20s internally).
- New `Backend.__healthcheck_readiness` coroutine: arm the healthcheck → await first `/health` 200 → benchmark → pubkey-wait → `_mark_loaded()`. `_start_tracking` runs it (plus log-tailing for `on_error`/`on_info` only) in healthcheck mode; the log `ModelLoaded` case is gated off.
- **Correctness invariants** (from adversarial review):
  - `errored ⟹ not loaded`: `__run_benchmark` calls `backend_errored` and returns `0.0` **without raising** on "no successful responses", so a readiness path could otherwise mark the model loaded on top of an error (contradictory loaded+errored, max_perf 0). A latched `__errored` flag + a single `_mark_loaded()` choke point enforce this across **all** readiness paths (log / lifecycle / healthcheck).
  - Health regressions are reported only once the model is **genuinely loaded** (`__model_loaded`), not merely after first-200. This closes a benchmark-window race and tolerates the connection-refused / 503-loading startup phase.

Log mode remains the **default** — `ace`/`wan`/`tgi`/`comfyui-json` keep log-based readiness. (The shared `__healthcheck` regression-gate change does mean log/lifecycle mode now reports a `/health` regression only after full load rather than after first-200 — marginally later crash detection during the load window, in exchange for the invariant.)

## Verified live

Both fixes and their tests are grounded in real instances (SGLang 0.5.15 srt + llama.cpp): both `/health` return **503 during model load** and **200 when busy**, and a live llama.cpp restart was observed transitioning `000 (conn refused) → 503 (loading) → 200 (ready)`. So `/health` is a valid readiness gate for both, and mid-load failures are expected (not regressions).

## Tests

14 tests in `tests/serverless/test_healthcheck_readiness.py`, using the repo's shared session-mock fixtures (`attach_serverless_backend_mock_aiohttp_session` + `make_serverless_aiohttp_get_context_manager`): config/thread-through, health→benchmark→loaded, timeout→errored, missing-url, mode validation, the race fix (health failure/conn-error before loaded does NOT error, after loaded DOES), and the errored⟹not-loaded fix (a failed benchmark must not mark loaded). Full serverless suite: **561 pass**.

## Known limitations / follow-ups (not in this PR)

- **`healthcheck` mode requires a readiness endpoint** (503-during-load), documented on the `readiness` field. A liveness-only `/health` (200 as soon as the socket binds) would trip readiness pre-load. Sharp edge, opt-in.
- The `openai` worker (pyworker repo) does **not** yet set `readiness='healthcheck'` — flipping it on, plus the `MODEL_LOG` env-drive, is the CON-1612 companion change.
- Pre-existing: the 60s pubkey-wait vs ~65s worst-case fetch (copied verbatim from the log path); `__read_logs` blocks if the log file never appears. Not touched here.
- Autoscaler-side: whether a set `error_msg` is treated as errored regardless of `model_is_loaded` is control-plane behavior outside this repo — worth confirming with that team.

### DIFF
--- a/vastai/serverless/server/lib/backend.py
+++ b/vastai/serverless/server/lib/backend.py
@@ -106,6 +106,14 @@ class Backend:
     session_metrics: Dict[str, RequestMetrics] = dataclasses.field(default_factory=dict)
     max_sessions: int = dataclasses.field(default=-1)
     lifecycle: Optional[AsyncContextManager] = dataclasses.field(default=None)
+    # Readiness mode: 'logs' (default — mark loaded on the on_load log line) or
+    # 'healthcheck' (mark loaded on the first successful /health probe, no log dep).
+    readiness: str = dataclasses.field(default="logs")
+    readiness_timeout: float = dataclasses.field(default=300.0)
+    # Per-probe HTTP timeout for the /health check (seconds). Some backends' health
+    # does real work (SGLang runs a 1-token gen, up to ~20s internally); set this >=
+    # the backend's own health timeout to avoid a false regression error on a stall.
+    healthcheck_probe_timeout: float = dataclasses.field(default=10.0)
 
     async def pyworker_update_handler(self, request: web.Request) -> web.Response:
         # Verify authorization header matches mtoken
@@ -354,6 +362,10 @@ class Backend:
         )
 
     def __post_init__(self):
+        if self.readiness not in ("logs", "healthcheck"):
+            raise ValueError(
+                f"invalid readiness={self.readiness!r}: expected 'logs' or 'healthcheck'"
+            )
         self.metrics = Metrics()
         self.metrics._set_version(self.version)
         self.metrics._set_mtoken(self.mtoken)
@@ -363,6 +375,16 @@ class Backend:
         self.__start_healthcheck: asyncio.Event = asyncio.Event()
         self.__healthcheck_ready: asyncio.Event = asyncio.Event()
         self.__healthcheck_succeeded: bool = False
+        # Set True once the model is GENUINELY loaded (after benchmark + pubkey), in
+        # any readiness path. The healthcheck loop reports regressions only once this
+        # is True, so a health blip during startup/benchmark can't error the backend
+        # (nor race _model_loaded). Verified against live SGLang + llama.cpp: both
+        # 503 during load, so first-200 == ready and mid-load failures are expected.
+        self.__model_loaded: bool = False
+        # Latched by backend_errored(). Enforces 'errored => not loaded': a failed
+        # benchmark calls backend_errored and (in the swallow path) returns 0.0 without
+        # raising, so a readiness path must NOT then mark the model loaded on top of it.
+        self.__errored: bool = False
 
     @cached_property
     def session(self):
@@ -606,7 +628,7 @@ class Backend:
 
         await self.__start_healthcheck.wait()
 
-        timeout = ClientTimeout(total=10)
+        timeout = ClientTimeout(total=self.healthcheck_probe_timeout)
 
         while True:
             try:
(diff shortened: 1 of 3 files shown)


vastai/serverless/server/lib/backend.py
```diff
@@ -637,13 +671,23 @@ async def __healthcheck(self) -> None:
...
             except Exception as e:
                 log.debug(f"Healthcheck failed with exception: {e}")
-                if self.__healthcheck_succeeded:
+                if self.__model_loaded:
                     self.backend_errored(str(e))
 
             await sleep(10)
 
     async def _start_tracking(self) -> None:
-        if self.lifecycle is not None:
```
LucasArmandVast, line 646: So with this proposed change, we are now adding a third way for the model to mark itself as ready. We have the original "logs" method, the "lifecycle" trigger for Deployments, and now the "healthcheck" trigger. I think that this is getting to the point that warrants a larger refactor to handle these three parallel readiness probes is worth investigating to maintain code cleanliness and clarity.

My initial thoughts on this: we should track a single, shared __readiness() asyncio coroutine that switches on the "self.readiness" flag. If "logs", wait for asyncio.Event emitted by __read_logs() when ModelLoaded log is loaded. If "lifecycle", just enter the current "__lifecycle_startup()" coroutine. If "healthcheck", wait for a asyncio.Event emitted by __healthcheck(). Then, there is only *one* `await gather(...` line, and readiness is consolidated into a single handler with three branches.

vastai/serverless/server/lib/backend.py
```diff
@@ -62,6 +62,13 @@
 SESSION_GC_INTERVAL = 5.0
 BENCHMARK_INDICATOR_FILE = ".has_benchmark"
 MAX_PUBKEY_FETCH_ATTEMPTS = 5
+# Secondary /health confirm timeout for the log and lifecycle readiness paths: those
+# already have a PRIMARY ready signal (the on_load log line / lifecycle __aenter__), so
+# this only bounds a health probe that never confirms afterward. Kept fixed and separate
+# from readiness_timeout — in 'healthcheck' mode readiness_timeout is the PRIMARY wait (a
+# large-model cold start) and is larger; coupling them would slow default-path failure
+# detection.
```
LucasArmandVast, line 70: Just a style nit, but I prefer explanatory LLM-generated comments like these be shortened to more concise, single-line descriptions (see above). Same for the other multi-line comments in this PR. I think most can be shortened or just removed entirely.

LucasArmandVast:
I think it makes sense to have a more generic readiness checker instead of tailing model logs. This is a much more user friendly and standard solution. What about for backends where healthy does not necessarily equal ready? For vLLM for example, is there a way to prove directly via API for if a given model is loaded? Is there a good generic solution for this?

robballantyne:
No worries on the healthcheck. We've dropped it from our side for now — the images just run log mode, so nothing's blocked. Happy to close this and revisit readiness when you do the consolidation.

PR #471, opened 2026-07-23
fix: keep API keys out of request URLs

### LINKED ISSUE #470: API key is included in request URLs
In short: [`VastClient._build_url()`](https://github.com/vast-ai/vast-cli/blob/v1.4.3/vastai/api/client.py#L55-L88) adds the API key as an `api_key` query parameter, even though `_build_headers()` also sends it as a `Bearer` token. This is despite the fact that the [authentication docs](https://docs.vast.ai/api-reference/authentication) specify header-only authentication.

This behaviour appears in all releases of `vastai` from 1.0.13 through 1.4.3, as well as the current `master`. You can observe this as follows:

```python
from vastai.api.client import VastClient

client = VastClient(
    api_key="REALLY_SHOULD_NOT_BE_HERE",
    server_url="https://console.vast.ai",
)

print(client._build_url("/instances"))
# https://console.vast.ai/api/v0/instances?api_key=REALLY_SHOULD_NOT_BE_HERE
```

This is an avenue for people to leak API keys into server logs, and anywhere else, since URLs are not intended for sensitive data and are not protected by HTTPS.

alisatwat3:
Stop VastClient and the deprecated vast.py CLI from adding API keys to URL query parameters. Authentication continues through the Authorization Bearer header.

Add regression tests ensuring URLs and caller-provided query arguments do not contain injected API keys.

Fixes #470

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -582,15 +582,13 @@ def apiurl(args: argparse.Namespace, subpath: str, query_args: Dict = None) -> s
 
     :param argparse.Namespace args: Namespace with many fields relevant to the endpoint.
     :param str subpath: added to end of URL to further specify endpoint.
-    :param typing.Dict query_args: specifics such as API key and search parameters that complete the URL.
+    :param typing.Dict query_args: search parameters that complete the URL.
     :rtype str:
     """
     result = None
 
     if query_args is None:
         query_args = {}
-    if args.api_key is not None:
-        query_args["api_key"] = args.api_key
     if not re.match(r"^/api/v(\d)+/", subpath):
         subpath = "/api/v0" + subpath
     
--- a/vast.py
+++ b/vast.py
@@ -336,7 +336,7 @@ class hidden_aliases(object):
         self.l.append(x)
 
 def http_request(verb, args, req_url, headers: dict[str, str] | None = None, json_data = None):
-    if headers is None:
+    if not headers:
         headers = apiheaders(args)
 
     t = 0.15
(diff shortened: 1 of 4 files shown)


vast.py
```diff
@@ -582,15 +582,13 @@ def apiurl(args: argparse.Namespace, subpath: str, query_args: Dict = None) -> s
...
     :param str subpath: added to end of URL to further specify endpoint.
-    :param typing.Dict query_args: specifics such as API key and search parameters that complete the URL.
+    :param typing.Dict query_args: search parameters that complete the URL.
     :rtype str:
     """
     result = None
 
     if query_args is None:
         query_args = {}
-    if args.api_key is not None:
```
vastzuby, line 592: so the `client.py` change works but not for `vast.py`

there's ~30 commands in `vast.py` with missing headers so they rely on the api key query param that this removes 

i know vast.py is deprecated but we still want to ensure backwards compatibility.

vastzuby: easiest fix is to revert the `vast.py` part 

or you could have http_request() fall back to apiheaders(args) when no headers get passed in, that covers all the call sites at once

otherwise the fix looks good 👍

alisatwat3: Image

is this fine? @vastzuby

vastzuby: i made one small change, otherwise looks good :+1:

vastzuby (approved):

vast.py
alisatwat3: thanks for accepting my contribution @vastzuby

PR #472, opened 2026-07-24
Add vastai uninstall command for managed installs

sammy-vastai:
## Summary
- Adds `vastai uninstall`, mirroring `vastai update`'s structure and same-method discipline: on a managed (curl | bash installer) install it removes the install root and its `~/.local/bin` symlinks while leaving `~/.config/vastai` (API key), cache, and state untouched; on a pip install it refuses and points at `pip uninstall vastai` instead of shelling out.
- Prompts for confirmation unless `-y`/`--yes`; refuses in a non-interactive shell without `--yes`.
- `perform_uninstall()` added to `selfupdate.py` next to the existing `perform_update()`.
- Both `update` and `uninstall` are hidden from `--help`/tab-completion for now (same `hidden=True` discoverability gate used for the network-volume commands in #466) — still fully runnable if typed directly, until they've had more real-world testing.

## Test plan
- [x] `poetry run pytest tests/cli/` — 389 passed
- [x] New `tests/cli/test_uninstall_command.py` covers: pip-install refusal, `--yes` skipping the prompt, non-interactive refusal without `--yes`, confirm/decline/EOF on the prompt, and `perform_uninstall()`'s filesystem behavior (removes root, removes only symlinks that resolve into the root, leaves foreign same-named binaries and config alone)
- [x] New hidden-command tests in `test_parser.py` confirm `update`/`uninstall` are excluded from `--help` but still parse
- [x] Manually verified end-to-end: built a dev wheel from this branch (`scripts/dev-install.sh --from-source`), confirmed `uninstall` is hidden from `--help` but runs directly, ran `vastai uninstall --yes` against a real managed install and confirmed the install root was removed and config was left in place, then reinstalled via `scripts/install.sh` and confirmed the wheel-install progress UX still works

### DIFF
--- a/tests/cli/test_parser.py
+++ b/tests/cli/test_parser.py
@@ -261,6 +261,15 @@ class TestFullCliHiddenCommands:
         args = cli_parser.parse_args(["search", "network-volumes"])
         assert callable(args.func)
 
+    def test_update_and_uninstall_are_hidden_from_help(self, cli_parser):
+        help_text = cli_parser.parser.format_help()
+        assert "\nupdate " not in help_text
+        assert "\nuninstall " not in help_text
+
+    def test_update_and_uninstall_still_parse_directly(self, cli_parser):
+        assert callable(cli_parser.parse_args(["update", "--check"]).func)
+        assert callable(cli_parser.parse_args(["uninstall", "--yes"]).func)
+
 
 class TestTwoStageCompletions:
     def setup_method(self):
--- a/tests/cli/test_uninstall_command.py
+++ b/tests/cli/test_uninstall_command.py
@@ -5,14 +5,16 @@ managed install root under tmp_path, argv-level tests via parse_argv, and
 the same-method discipline (pip installs are refused, never shelled out to).
 """
 
+import shutil
 import sys
+from pathlib import Path
 from types import SimpleNamespace
 from unittest.mock import patch
 
 import pytest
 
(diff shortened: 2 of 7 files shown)


sammy-vastai:
Addressed Copilot's review:

- `perform_uninstall()` no longer silently swallows failures: it resolves `install_root()` before comparing against resolved symlink targets (fixes the mismatch when `VASTAI_INSTALL_DIR` is itself a symlink) and now raises a new `UninstallError` with the real `OSError` detail on any removal failure, instead of `ignore_errors=True`/bare `except: pass`.
- `vastai uninstall` now treats removal as fallible: it catches `UninstallError`, prints the real error to stderr, and exits 1 instead of unconditionally printing success. The success message now uses the `Path` returned by `perform_uninstall()`.
- Replaced the hardcoded `~/.config/vastai` in the command's epilog/messages and the `perform_uninstall()` docstring with a reference to `vastai.cli.util.DIRS` (XDG-resolved at runtime), so the text stays correct under `$XDG_CONFIG_HOME` overrides.
- Added regression tests: a symlinked `VASTAI_INSTALL_DIR`, symlink-unlink failure, root-removal failure, and the command surfacing a `perform_uninstall()` failure as a non-zero exit with the error on stderr.

`tests/cli/` — 394 passed.

vastzuby (approved):

PR #475, opened 2026-07-25
Hide unreleased network-disk commands from --help

sammy-vastai:
## Summary

Follow-up to #466 (network volumes). Network disks aren't released either — `show network-disks` and `add network-disk` were still fully visible in `vastai --help` and tab completion.

Reuses the same `HIDDEN_COMMANDS` mechanism from #466: discoverability gate only, not an access gate — both commands still run if typed directly.

## Test plan
- [x] `poetry run pytest tests/cli` — 415 passed
- [x] Manually verified: neither command appears anywhere in `vastai --help`; both still parse and run when typed directly

### DIFF
--- a/tests/cli/test_parser.py
+++ b/tests/cli/test_parser.py
@@ -210,6 +210,8 @@ class TestIsHiddenCommand:
         assert is_hidden_command("create network-volume") is True
         assert is_hidden_command("list network-volume") is True
         assert is_hidden_command("unlist network-volume") is True
+        assert is_hidden_command("show network-disks") is True
+        assert is_hidden_command("add network-disk") is True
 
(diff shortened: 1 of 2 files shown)


vastzuby (approved):

vastzuby:
imo should just remove everything related to network volumes, we can always add it from the history

sammy-vastai:
> imo should just remove everything related to network volumes, we can always add it from the history

Yeah didn't know whether to delete so that's why I tagged some infra/host devs.  I think everything network volumes related is now hidden (from --help and autocomplete suggestions).  Last pr removed network volume commands and i had missed the 'network disk' related commands.  The commands still work if we type them, they're just hidden from users.  Should be sufficient, lmk if any concerns

PR #476, opened 2026-07-27
installer: show a progress bar with one live log line, not scrolling uv output

sammy-vastai:
#467 dropped `--quiet` from the wheel install so a TTY install showed progress. It showed too much: the CPython download plus ~40 `+ pkg==ver` lines scroll the "Get started" block off the screen, which is the one thing a first-time installer needs to read (a coworker missed it entirely).

At a TTY the install now renders as a fixed two-line region on stderr — a step bar plus one live log line that tails whatever `uv` is doing:

```
  Installing vastai 1.5.0 (Python 3.12) → /home/you/.local/share/vastai
  ⠹ ━━━━━━━━━━━━━━━───────────  60%  Installing vastai 1.5.0
    ↳  + cryptography==49.0.0
```

The region is erased when the install finishes, so the last thing on screen is the next-steps block. A full install leaves ~14 lines instead of ~60.

## How it works

- A background ticker animates the spinner; steps and tool output reach it through state files under the temp workdir. Both writers repaint with a single `printf`, so frames can't interleave, and a state change repaints immediately — a step that finishes inside a tick still shows.
- `ui_run` tails a command's output through `tee`; on failure `ui_fail` tears the region down and replays the tail, so `uv`'s own diagnostic ("No solution found when resolving dependencies…") is still what the user sees. No `tr`/`sed` in that pipe: they block-buffer into a pipe and hold every line back until the command exits.
- Widths come from `/dev/tty` first (`curl | bash` leaves stdin as the pipe), falling through to terminfo, with a sized-0 pty rejected. Nothing exceeds the terminal width. ASCII glyphs (`#`/`.`/`-\|/`) replace the Unicode ones outside UTF-8 locales.
- `INT`/`TERM`/`HUP` are trapped so Ctrl-C erases the region and restores the cursor instead of leaving it hidden.

## Unchanged without a TTY

CI and pipes get plain step lines and tool output straight through, no escape codes — `uv python install` keeps `--quiet` there as before. `VASTAI_PROGRESS=0|1` forces either mode.

## Testing

- 18/18 hermetic installer tests pass, 3 new: bar rendering with the hints surviving, the `VASTAI_PROGRESS=0` opt-out, and failure diagnostics surviving the region. The fake `uv` now emits real-shaped chatter and can fail on demand. A `run_install_tty` helper replaces the pty invocation duplicated across the three existing pty tests. `fresh_install` also asserts no escape codes when detached from a tty.
- `shellcheck -S style` clean.
- Real end-to-end installs against the production manifest, including through `curl … | bash`, at 45/50/80/100 columns, in a UTF-8 and an `LC_ALL=C` locale, plus a forced pip failure and a Ctrl-C mid-install.

[Screencast from 2026-07-27 10-45-26.webm](https://github.com/user-attachments/assets/57625d2d-93c9-4957-a780-a1139db44071)

### DIFF
--- a/docs/install-design.md
+++ b/docs/install-design.md
@@ -75,6 +75,26 @@ curl -fsSL https://vast.ai/install.sh | bash
    immediately (no PyPI index propagation window) — into an isolated venv.
    **System Python is never used, even if present.**
 
+### Install output at a terminal
+
+The bootstrap tools are chatty — a 33 MB CPython download, then ~40 `+ pkg==ver`
+lines — and that scrolled the "Get started" block off the screen, which is the
+one thing a first-time installer needs to read. So at a TTY the install renders
+as a fixed two-line region on stderr: a step bar plus a single live log line
+that tails whatever `uv` is doing (`Downloading cpython-… (32.6MiB)`, then
+package names). The region is erased when the install finishes, leaving ~9
+lines total, ending with the next steps.
+
+- A background ticker animates the spinner; steps and tool output reach it
+  through state files under the temp workdir, so a step change repaints
+  immediately and the two writers can never split a frame.
+- Tool output is tailed through `tee`, with the full copy replayed if that step
+  fails — a wrapped command's own diagnostics are the real error, so they must
+  outlive the region.
+- No TTY (CI, pipes) → plain step lines and tool output straight through, as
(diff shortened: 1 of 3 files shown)


vastzuby (approved):

vastzuby:
looks sick

PR #484, opened 2026-07-31
Client/host command classification with role-aware CLI display + auto-detection (CLN-3582)

sammy-vastai:
Renters (clients) no longer see host-only commands in `--help`/tab completion (machine management, listings, min-bid, maintenance, etc.); hosts see everything. Display-only — the server still enforces real permissions, so any command still runs if typed directly.

## How
- Every command is classified host-only or not (`is_host_only_command()` in `vastai/cli/parser.py`) via an explicit `host_only=` kwarg, a per-command override set, or a per-module default (`machines.py`, `metrics.py`). Unclassified defaults to visible.
- Role (`client`/`host`) is stored in `~/.config/vastai/vast_role`. Unset = client, the default.
- Auto-detected lazily on the user's next real command (`get_client()` → `ensure_host_role_detected()`), via `GET /users/current`'s `host_agreement_accepted` flag — resolves correctly even before a new host's first machine listing. Cached permanently once resolved; `vastai set role host|client` overrides manually.
- The existing hand-written `[Host]` help-text prefixes are preserved as-is — no auto-generated prefix.

## Validation
- `poetry run pytest tests/cli` — 458 passed
- `bash tests/installer/run_tests.sh` — 19 passed
- CI green end to end

[CLN-3582](https://vastai.atlassian.net/browse/CLN-3582)

### DIFF
--- a/tests/cli/test_auth_commands.py
+++ b/tests/cli/test_auth_commands.py
@@ -94,6 +94,7 @@ class TestSetApiKey:
         legacy_file = tmp_path / ".vast_api_key"
         monkeypatch.setattr("vastai.cli.util.APIKEY_FILE", str(key_file))
         monkeypatch.setattr("vastai.cli.util.APIKEY_FILE_HOME", str(legacy_file))
+        monkeypatch.setattr("vastai.api.machines.show_machines", lambda client: [])
         from vastai.cli.commands.auth import set__api_key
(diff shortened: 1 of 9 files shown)


sammy-vastai:
https://claude.ai/code/artifact/128fbf41-5e37-4bb4-b4ee-36fa2dbc7c8e?via=auto_preview

vastzuby (approved):

PR #492, opened 2026-08-12
fix(cli): prompt to set api key on 401 when none is configured (CLN-3748)

sammy-vastai:
Jira: [CLN-3748](https://vastai.atlassian.net/browse/CLN-3748)

## Summary
- When a request fails with 401 and neither `$VAST_API_KEY` nor the saved key file has a key, print a clear prompt (`vastai set api-key <KEY>` + link to create one) instead of a bare `Failed with error 401: ...`.
- Named the "neither key source present" condition as `key_missing` for readability.

## Test plan
- [x] Added `test_401_with_no_api_key_configured_prompts_to_set_one` covering the new message.
- [x] `poetry run pytest tests/cli/` — 432 passed.

### DIFF
--- a/vastai/cli/main.py
+++ b/vastai/cli/main.py
@@ -68,7 +68,13 @@ def _emit_error(args, status_code, message):
             except OSError:
                 pass
 
-        if env and file_key:
+        key_missing = not env and not file_key
+
+        if key_missing:
+            print("  No API key is configured.", file=sys.stderr)
+            print("  Run: vastai set api-key <KEY>", file=sys.stderr)
+            print("  Create a key at https://console.vast.ai/manage-keys/", file=sys.stderr)
+        elif env and file_key:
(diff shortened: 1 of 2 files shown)


Marcoh572 (approved):

vastzuby (approved):

vastai/cli/main.py
```diff
@@ -68,7 +68,13 @@ def _emit_error(args, status_code, message):
             except OSError:
                 pass
 
-        if env and file_key:
+        key_missing = not env and not file_key
+
+        if key_missing:
+            print("  No API key is configured.", file=sys.stderr)
+            print("  Run: vastai set api-key <KEY>", file=sys.stderr)
+            print("  Create a key at https://console.vast.ai/manage-keys/", file=sys.stderr)
```
robertfernandez-vast, line 76: This checks whether the `VAST_API_KEY` environment variable exists or if the api key file exists and isn't empty. It doesn't try to use the api key to see if it works. If someone has that variable set but incorrectly or if the file has bad contents, this won't catch it.

PR #494, opened 2026-08-14
docs(cli): document host_id as a search offers filter field

guthrie-vast:
## What

Adds `host_id` to the **Available fields** list in `vastai search offers --help`, and adds an example showing how to exclude specific hosts with `notin`.

## Why

`host_id` has always been filterable in offer search — including `host_id notin [...]` for excluding hosts — but it was absent from the help's field list (while `machine_id` was listed), so users had no way to discover it. Verified against the live API: `notin` excludes exactly the listed hosts for both authenticated and anonymous queries.

No behavior change — help text only.

### DIFF
--- a/vast.py
+++ b/vast.py
@@ -4270,6 +4270,9 @@ def search__invoices(args):
             # search for reliable 4 gpu offers in Taiwan or Sweden
             vastai search offers 'reliability>0.99 num_gpus=4 geolocation in [TW,SE]'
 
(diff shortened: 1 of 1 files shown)


vastzuby (approved):

vastzuby:
`vast.py` is deprecated btw

vastzuby (changes requested):
Need to also update `vastai/cli/commands/offers.py` with same change since `vast.py` is deprecated.

PR #497, opened 2026-08-25
SDK: fix the parameters the docs promise but the code rejects, drops, or mangles

vastzuby:
A customer read `jupyter` off the published SDK reference, called `VastAI.create_instance(jupyter=True)`, and got a `TypeError`. They said there were "many discrepancies". They were right, and it was worse than the crash they hit.

Four separate failure modes, all from the same root cause: **50 documented parameters that don't exist**, parameters that bind and then get sent in the wrong shape, parameters that are accepted and silently thrown away, and malformed dates that quietly return data for a range you didn't ask for. The loud one was the only one anybody could see.

## The bugs

**1. Documented parameters that raise.** `create_instance`, `create_instances`, `launch_instance`, `update_template`, `cloud_copy` and three others publish parameters the method rejects outright.

**2. Parameters accepted but sent wrong.** `env` is documented as a string because the CLI flag is a string, but the wire wants a dict. The CLI runs `parse_env()`; the SDK shipped the raw string. Same shape for `args`, and for five search methods that forwarded a CLI-style query string into a layer contracted on dicts, so the filter silently did nothing.

**3. Parameters accepted and thrown away.** `create_template` popped `search_params` and `no_default` and discarded them. A test asserted this was correct.

**4. Bad dates return plausible wrong data.** `show_invoices` / `show_earnings` swallowed a `ValueError` on an unparseable date and silently substituted yesterday.

## Why they exist

`VastAI.create_instance` took `**kwargs` and forwarded to a delegate with a closed signature, so the wrapper advertised a surface the delegate did not have.

The docs generator builds the parameter table for any `**kwargs` method by scraping the matching CLI command's argparse flags. `--jupyter`, `--onstart` and `--entrypoint` are CLI-only concepts — the CLI collapses them into a `runtype` string before the API layer is reached — but they got published as SDK parameters anyway.

The drift checker is told not to judge extra documented parameters on `**kwargs` methods, so nothing caught it.

## The fix

**One shared payload builder.** `build_create_instance_payload()` in `vastai/api/instances.py` is now the single place the CLI, the SDK and the docs generator agree on what a create-instance call takes. The CLI's `get_runtype` / `validate_volume_params` / `validate_portal_config` became exception-raising library functions; the argparse-coupled duplicates are deleted, along with duplicate copies of `parse_env` and `smart_split`.

**Closed signatures.** `create_instance`, `create_instances`, `launch_instance`, `update_template`, `cloud_copy`, `search_offers`, `show_invoices` and `show_earnings` no longer take `**kwargs`, so the generator introspects them instead of guessing. The docs become correct by construction. This also means the drift checker starts judging them — no change to the checker was needed.

**Query strings parse.** The five search methods now parse a CLI-style string into the filter dict. Dicts still pass through untouched.

**Bad dates raise**, naming the argument and the expected format.

## Backwards compatibility

Closed signatures stop accepting what they do not name, so I diffed the accepted-keyword surface of every `VastAI` method against master. It found two breaks, both fixed here:

- `update_template` used to forward `**kwargs` straight through, so the raw api names (`runtype`, `use_ssh`, `extra_filters`, `private`, …) are what every working script uses today — the friendly names all raised. Both spellings work now, raw wins.
- `create_template` without `search_params` would have started attaching three default offer filters where it previously sent `{}`.

New parameters are appended, never inserted, so positional calls bind the same values. `tests/sdk/test_backwards_compatibility.py` pins it.

**One deliberate exception: `disable_bundling` is removed from `search_offers` / `search_offers_new`.** It was reachable through `**kwargs` before and did work, but CLN-2792 removed the `--disable-bundling` CLI flag and scrubbed it from the OpenAPI specs, leaving the SDK as the only surface still carrying it. Closing these signatures would have promoted it from an invisible passthrough to a published parameter, so it comes out instead. The api layer keeps the plumbing for internal use. This is the only call shape in this PR that worked before and does not now.

**The `runtype` tri-state is deliberate.** A bare CLI create sends `runtype: "ssh"`; the SDK sends no `runtype` key at all. Adopting the CLI default would silently change the payload for every existing SDK caller. Pinned by a test.

## Testing

- **New CI test** (`tests/scripts/test_sdk_documented_params.py`) binds every generator-produced parameter against the real method with the transport stubbed, and asserts it reaches the payload — so a future fix can't turn rejections into silent drops. **50 rejected → 0.**
- Suite: 1469 passed. The 16 failures / 12 errors match the master baseline exactly (pre-existing API-key and tarball tests).
- Payload differential, CLI vs SDK on identical input: identical on every case except the intentional `runtype` tri-state.
- **Live wire test.** Rented real instances and destroyed them. The customer's exact call now produces `image_runtype = 'jupyter_direc ssh_direc ssh_proxy'` and `status_msg = success, running .../jupyter`. Bulk create, `launch_instance`, template create/update/delete, and the five search methods were all exercised against the live API.

## Deferred

- 23 SDK methods still have no reference page.
- `cloud_copy` scheduling needs `add_scheduled_job` exposed as an SDK method. The docs no longer advertise it.
- The checker cannot see through the docs repo's snippet-import pages, so parameters documented there are invisible to it. Pre-existing.
- 32 parameters publish with an empty description. The generator sources per-parameter help from CLI flags only, so a real signature parameter with no matching flag gets nothing — `runtype`, `extra`, `volume_info`, the raw template fields. Master is at 59 empty of 449 and this branch at 78 of 447, so closing the signatures surfaced more of them. The fix is a fallback to the docstring `Args:` block in the generator, which would also clear most of the pre-existing ones. Deliberately left for a follow-up rather than widening this PR into the generator.


Gobind-Vast:
### Code review

Found 1 issue:

1. Closing the `search_offers` / `search_offers_new` signatures newly publishes the deprecated `disable_bundling` flag to the generated SDK reference, described as a live feature. `vastai/api/offers.py` calls the same parameter "Deprecated bundling flag", and cb65d799 ("CLN-2792: remove disable-bundling references from CLI and API docs") deliberately removed the `--disable-bundling` CLI flag and scrubbed it from the OpenAPI specs — there is no `disable_bundling` anywhere under `vastai/cli/` today. Because the generator only fabricates parameters from CLI flags for `**kwargs` methods and otherwise introspects the real signature, `extract_sdk_methods` now emits `disable_bundling` for both methods where it previously emitted none (verified by running the generator against both this head and `38e0711`). Worth either keeping it out of the public signature or marking it deprecated in the docstring, as `show_invoices` already does.

https://github.com/vast-ai/vast-cli/blob/d033b9529779b1bac9be09be8aefa831e400992b/vastai/sdk.py#L276-L290

https://github.com/vast-ai/vast-cli/blob/d033b9529779b1bac9be09be8aefa831e400992b/vastai/api/offers.py#L22-L25

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vastzuby:
@Gobind-Vast  fixed the disable bundling issue

Gobind-Vast:
Confirmed fixed at 4457a3f — `extract_sdk_methods` now emits `['query', 'type', 'order', 'limit', 'storage', 'no_default']` for both methods, identical to `38e0711`, and the api-layer plumbing is retained. Thanks.

One thing worth a line in the description: dropping the parameter from `SyncClient.search` / `AsyncClient.search` is a hard compat break, not just a docs change. On `VastAI.search_offers` the flag was only reachable through `**kwargs` and was never documented, so nothing was promised. On the two clients it was an explicitly named keyword parameter, and both are exported in `__all__` and documented in `vastai_sdk/SKILL.md`, so `SyncClient.search(disable_bundling=True)` goes from working to `TypeError`.

https://github.com/vast-ai/vast-cli/blob/38e0711b4f5d946484147cb9e5e3d76a49db12f8/vastai/sync/client.py#L32-L34

https://github.com/vast-ai/vast-cli/blob/4457a3f66bf318f4a5941eef904758ed7b619fad/vastai/sync/client.py#L31-L34

Defensible given the flag is deprecated and the goal was consistency across surfaces — and the new test says so explicitly. But the backwards-compatibility section only covers the accepted-keyword surface of `VastAI` methods, so this break currently isn't recorded anywhere a caller would look.

🤖 Generated with [Claude Code](https://claude.ai/code)

Gobind-Vast (approved):

PR #498, opened 2026-08-27
fix(cli): resolve .pub file paths in `create ssh-key`

Kilo-Loco:
## The bug

`create ssh-key` imports `get_ssh_key` and then never calls it — it passes `args.ssh_key` straight to the API:

```python
from vastai.cli.util import get_ssh_key, generate_ssh_key   # imported, unused

ssh_key_content = args.ssh_key
...
result = keys_api.create_ssh_key(client, ssh_key=ssh_key_content)
```

So running the form the [hello-world guide](https://docs.vast.ai/cli/hello-world) and the command's own epilog prescribe:

```
vastai create ssh-key ~/.ssh/id_ed25519.pub
```

stores the literal string `/home/you/.ssh/id_ed25519.pub` as the account's public key:

```
$ vastai show ssh-keys
[{'id': 1276050, ..., 'public_key': '/Users/me/.ssh/id_ed25519.pub', ...}]
```

The command prints success. The key is silently useless. The first symptom is `Permission denied (publickey)` on an instance that has already been created and is already billing.

`update ssh-key` and `attach ssh` both call `get_ssh_key`; only `create` misses it — so the one broken path is the account-level one that new users are walked through first. This restores the behavior originally requested in #133.

## The fix

One line, routing through the existing helper:

```python
ssh_key_content = get_ssh_key(ssh_key_content)
```

That also picks up `get_ssh_key`'s two guards: a private key or a non-key string now raises before anything is sent, rather than being stored verbatim.

## Why it wasn't caught

The existing test asserted only that the parameter was present:

```python
assert "ssh_key" in call_args[1]["json_data"]
```

which holds for any value, including a file path. It now asserts the key material itself, plus three new cases: a `.pub` path is read, a private key is rejected, a non-key string is rejected. All three fail on `master` and pass with the fix.

## Verification

```
$ pytest tests/cli/test_keys_commands.py -q     # before fix
3 failed, 6 passed

$ pytest tests/cli/test_keys_commands.py -q     # after fix
9 passed

$ pytest tests/cli tests/sdk -q
531 passed
```

Full `pytest tests/` shows 12 failed / 12 errors both with and without this change (tar-utils and credential-requiring suites) — identical on clean `master`, unrelated to this PR.

## Not addressed here

`get_ssh_key` uses `os.path.exists` without `expanduser`, so a *quoted* `"~/.ssh/id_ed25519.pub"` still falls through as a literal string. Unquoted paths are shell-expanded and unaffected. That's a shared helper touching `update`/`attach` too, so I left it out to keep this focused — happy to follow up if you want it.

🤖 Generated with [Claude Code](https://claude.com/claude-code) - And reviewed by Kyle's eyes

### DIFF
--- a/tests/cli/test_keys_commands.py
+++ b/tests/cli/test_keys_commands.py
@@ -38,15 +38,60 @@ class TestShowApiKey:
         assert "/auth/apikeys/1/" in call_args[0][0]
 
 
+PUBLIC_KEY = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIExampleKeyMaterial user@host"
+
+PRIVATE_KEY = """-----BEGIN OPENSSH PRIVATE KEY-----
+b3BlbnNzaC1rZXktdjEAAAAABG5vbmUAAAAEbm9uZQAAAAAAAAABAAAAMwAAAAtzc2gtZW
+-----END OPENSSH PRIVATE KEY-----
+"""
+
+
 class TestCreateSshKey:
     def test_create_ssh_key(self, parse_argv, patch_get_client, mock_response, capsys):
         patch_get_client.post.return_value = mock_response(200, {"id": 1, "success": True})
-        args = parse_argv(["create", "ssh-key", "ssh-rsa AAAA... user@host"])
+        args = parse_argv(["create", "ssh-key", PUBLIC_KEY])
         args.func(args)
         patch_get_client.post.assert_called_once()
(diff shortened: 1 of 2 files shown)


Kilo-Loco:
Closing this for now — opened before the author had reviewed it. Will reopen once verified.

Kilo-Loco:
Superseded by #499, which carries the same fix plus the follow-up review changes (clean error on an unreadable path, tighter assertions) in a single commit.

PR #500, opened 2026-08-27
fix(AUTO-1898): drop endpoint-only target_util and cold_mult from workergroup

madisonvast:
## AUTO-1898: drop endpoint-only `target_util` and `cold_mult` from the workergroup CLI

`--target_util` and `--cold_mult` are endpoint-level scaling knobs. The workergroup
commands have accepted them for a long time, and `create workergroup --help` even
carried a `[NOTE: this field isn't currently used at the workergroup level]` on each.
This removes them so the flags live only where they do something.

### Why they're dead

The autoscaler parses `target_util`/`cold_mult` onto `Autogroup` (`autoscaler.cpp`
`parse_group`) and then never reads them — they're echoed back in sim serialization and
settable via scenario events, and that's it. The only consumer of these values is
`asm_ratio_manager.cpp`, where the group is an `Endptgroup`. `benchmarks.py` already
skips them for exactly this reason ("the autoscaler reads both from the endpoint group,
not the workergroup — verified via autoscaler.cpp"), and the web UI already strips them
from workergroup payloads.

### Changes

- `vastai/cli/commands/endpoints.py` — remove both flags from `create workergroup` and
  `update workergroup` and from the `endpoints_api` call sites; each epilog now says
  where the knobs actually live.
- `vastai/api/endpoints.py` — remove both keys from the `create_workergroup` /
  `update_workergroup` request bodies and from the docstring args. Endpoint functions
  untouched.
- `vast.py` — same removal in the deprecated monolith, so the two clients agree.
- `tests/cli/test_endpoints_commands.py` — new coverage: both flags rejected by both
  workergroup commands, neither key present in the POST/PUT body, and both still
  accepted by `create endpoint` / `update endpoint`.

### Backend impact: none

No server change is needed and no coordination is required in either direction. The
workergroup param models are `extra='ignore'`, so older CLIs keep working, and the
payload previously sent `"target_util": None` whenever the flag was unused — which the
server already treats as "inherit from the endpoint". For every call that wasn't passing
the flags, the wire format is unchanged.

### Behavior change

`create workergroup --endpoint_name <name>` for an endpoint that doesn't exist yet makes
the server create it, and it used to seed that new endpoint with these values. Such
endpoints now get the server defaults (`target_util` 0.9, `cold_mult` 3.0). Create the
endpoint explicitly first if you need non-defaults.

### Breaking

`vastai create|update workergroup --target_util X` (or `--cold_mult`) now exits 2 with
"unrecognized arguments". The values were being discarded by the autoscaler anyway, so
no scaling behavior changes for anyone — except the endpoint-creation case above.

### Out of scope

- `WorkergroupConfig` still declares both fields, so the async serverless client and
  `VastAI().create_workergroup(target_util=...)` still send them (server ignores them).
  Worth a follow-up so the SDK stops silently swallowing the value.
- `min_load` is equally inert at the workergroup level and is left as-is.
- Removing the `autoscale_jobs` columns / the backend params is a separate cleanup
  (migration + autoscaler + openapi snapshot).

### Testing

- `pytest tests/cli tests/test_legacy_vast.py tests/test_data_objects.py tests/scripts`
  → 606 passed.
- Full suite matches its pre-change baseline exactly (365 pre-existing failures from a
  venv without `pytest-asyncio`/`pyOpenSSL`), plus the 8 new tests.
- `vastai create workergroup --help` and `vastai update workergroup --help` verified.

### Follow-up chores

- Regenerate the `vast-ai/docs` pages for these two commands; `verify_cli_sdk_docs.py
  --check-params` will report parameter drift until then.
- `tests/compat_test.sh` will show an expected `--help` DIFF for both commands.

### DIFF
--- a/vastai/cli/commands/endpoints.py
+++ b/vastai/cli/commands/endpoints.py
@@ -222,14 +222,15 @@ def get__endpt_workers(args):
     argument("--gpu_ram",     help="estimated GPU RAM req  (independent of search string)", type=float),
     argument("--search_params", help="search param string for search offers    ex: \"gpu_ram>=23 num_gpus=2 gpu_name=RTX_4090 inet_down>200 direct_port_count>2 disk_space>=64\"", type=str),
     argument("--min_load", help="[NOTE: this field isn't currently used at the workergroup level] minimum floor load in perf units/s  (token/s for LLms)", type=float),
-    argument("--target_util", help="[NOTE: this field isn't currently used at the workergroup level] target capacity utilization (fraction, max 1.0, default 0.9)", type=float),
-    argument("--cold_mult",   help="[NOTE: this field isn't currently used at the workergroup level]cold/stopped instance capacity target as multiple of hot capacity target (default 2.0)", type=float),
     argument("--cold_workers",   help="min number of workers to keep 'cold' for this workergroup", type=int),
     argument("--auto_instance", help=argparse.SUPPRESS, type=str, default="prod"),
     usage="vastai create workergroup [OPTIONS]",
     help="Create a new autoscale group",
     epilog=deindent("""
         Create a new autoscaling group to manage a pool of worker instances.
+
+        target_util and cold_mult are endpoint-level; set them with 'vastai create endpoint'
+        or 'vastai update endpoint'.
 
         Example: vastai create workergroup --template_hash HASH  --endpoint_name "LLama" --test_workers 5
     """),
@@ -248,7 +249,6 @@ def create__workergroup(args):
             endpoint_name=args.endpoint_name, endpoint_id=args.endpoint_id,
             test_workers=args.test_workers, gpu_ram=args.gpu_ram,
             search_params=args.search_params, min_load=args.min_load,
-            target_util=args.target_util, cold_mult=args.cold_mult,
             cold_workers=args.cold_workers, auto_instance=args.auto_instance,
         )
         print("workergroup create {}".format(result))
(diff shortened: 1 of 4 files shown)


vastai/cli/commands/endpoints.py
```diff
@@ -222,15 +222,16 @@ def get__endpt_workers(args):
...
     argument("--cold_workers",   help="min number of workers to keep 'cold' for this workergroup", type=int),
     argument("--auto_instance", help=argparse.SUPPRESS, type=str, default="prod"),
     usage="vastai create workergroup [OPTIONS]",
     help="Create a new autoscale group",
     epilog=deindent("""
         Create a new autoscaling group to manage a pool of worker instances.
 
+        target_util and cold_mult are endpoint-level; set them with 'vastai create endpoint'
+        or 'vastai update endpoint'.
+
```
LucasArmandVast: Don't need this warning

vastai/cli/commands/endpoints.py
```diff
@@ -300,16 +298,17 @@ def show__workergroups(args):
     usage="vastai update workergroup WORKERGROUP_ID --endpoint_id ENDPOINT_ID [options]",
     help="Update an existing autoscale group",
     epilog=deindent("""
-        Example: vastai update workergroup 4242 --min_load 100 --target_util 0.9 --cold_mult 2.0 --search_params \"gpu_ram>=23 num_gpus=2 gpu_name=RTX_4090 inet_down>200 direct_port_count>2 disk_space>=64\" --launch_args \"--onstart onstart_wget.sh  --env '-e ONSTART_PATH=https://s3.amazonaws.com/public.vast.ai/onstart_OOBA.sh' --image atinoda/text-generation-webui:default-nightly --disk 64\" --gpu_ram 32.0 --endpoint_name "LLama" --endpoint_id 2
+        target_util and cold_mult are endpoint-level; set them with 'vastai update endpoint'.
```
LucasArmandVast: Same here, not needed

LucasArmandVast (approved):

PR #501, opened 2026-08-28
Fix/auto 1912 remove test workers etc from cli

madisonvast:
## AUTO-1912: drop dead `test_workers`, `cold_workers`, and `min_load` from the workergroup CLI

CLI-only slice of AUTO-1909. `vastai create workergroup` and `vastai update workergroup` exposed `--test_workers`, `--cold_workers`, and `--min_load`, none of which do anything at the workergroup level. Advertising them invites users to tune knobs that have no effect. Same treatment AUTO-1898 (#500) gave `--target_util` and `--cold_mult`.

No webserver or autoscaler change. No UI or docs-site change; those are separate slices of AUTO-1909.

### Why these are dead

Checked against the autoscaler source, not the help text:

- **`min_load`** is parsed into `Autogroup::min_load_` (`autoscaler.cpp:714`) and never read. The only consumer is `asm_ratio_manager.cpp:620-625`, which takes an `Endptgroup` (it touches `min_cold_load_`, a field `Autogroup` doesn't have).
- **`cold_workers`** isn't even parsed for `Autogroup`. `autoscaler.cpp:749` is inside the `Endptgroup` overload of `parse_group`; the `Autogroup` overload has no such line. Only sim/scenario code ever sets `Autogroup::cold_workers_`.
- **`test_workers`** does not appear in the autoscaler at all. The backend writes it to `autoscale_jobs.test_workers` and echoes it in `show workergroups`; nothing reads it.

### Changes

| File | What |
|---|---|
| `vastai/cli/commands/endpoints.py` | Three flags and their call-site kwargs removed from `create__workergroup` and `update__workergroup`; both epilog examples stripped of `--test_workers 5` / `--min_load 100` |
| `vastai/api/endpoints.py` | Three keys removed from the `create_workergroup` / `update_workergroup` `json_blob`s and from the `create_workergroup` docstring |
| `vast.py` | Same removals mirrored in the deprecated monolith, plus a commented-out `json_blob` that referenced three now-nonexistent `args.*` |
| `vastai/data/workergroup.py` | `WorkergroupConfig` loses all five fields: these three plus `target_util` / `cold_mult`, which AUTO-1898 deliberately left behind |
| `tests/cli/test_endpoints_commands.py` | `TestWorkergroupEndpointOnlyScalingFlags` → `TestWorkergroupDeadScalingFlags`, parameterized over all five |

`create_endpoint` / `update_endpoint` and their identically named keys are untouched. `sdk.py` needed no change; both workergroup wrappers are `**kwargs` passthroughs.

### Wire compatibility

Non-breaking for existing callers. The CLI already sent these as `None` unless explicitly passed, and the backend treats `None` as inherit-from-endpoint. The workergroup pydantic models are `extra='ignore'`, so dropping the keys entirely needs no backend coordination.

### DIFF
--- a/vastai/data/workergroup.py
+++ b/vastai/data/workergroup.py
@@ -18,13 +18,8 @@ class WorkergroupConfig:
     template_hash: Optional[str] = None
     template_id: Optional[int] = None
     launch_args: Optional[str] = None
-    min_load: Optional[float] = None
     min_cold_load: Optional[float] = None
-    target_util: Optional[float] = None
-    cold_mult: Optional[float] = None
-    cold_workers: Optional[int] = None
     max_workers: Optional[int] = None
-    test_workers: Optional[int] = None
     gpu_ram: Optional[float] = None
     autoscaler_instance: Optional[str] = None
     docker_login_user: Optional[str] = None
--- a/tests/cli/test_endpoints_commands.py
+++ b/tests/cli/test_endpoints_commands.py
@@ -49,21 +49,34 @@ class TestShowWorkergroups:
         assert "/autojobs/" in call_args[0][0]
 
 
-ENDPOINT_ONLY_SCALING_FLAGS = ("--target_util", "--cold_mult")
+WORKERGROUP_DEAD_FLAGS = (
+    ("--target_util", "0.9"),
+    ("--cold_mult", "2.5"),
+    ("--min_load", "100"),
+    ("--cold_workers", "5"),
+    ("--test_workers", "3"),
+)
+
+ENDPOINT_SCALING_FLAGS = (
+    ("--target_util", "0.9", 0.9),
+    ("--cold_mult", "2.5", 2.5),
+    ("--min_load", "100", 100.0),
+    ("--cold_workers", "5", 5),
+)
 
 
-class TestWorkergroupEndpointOnlyScalingFlags:
-    """target_util and cold_mult scale the endpoint group, never a workergroup."""
+class TestWorkergroupDeadScalingFlags:
+    """These scale the endpoint group or nothing at all; a workergroup never reads them."""
 
-    @pytest.mark.parametrize("flag", ENDPOINT_ONLY_SCALING_FLAGS)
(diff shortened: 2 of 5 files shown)


vastai/data/workergroup.py
```diff
@@ -18,13 +18,8 @@ class WorkergroupConfig:
     template_hash: Optional[str] = None
     template_id: Optional[int] = None
     launch_args: Optional[str] = None
-    min_load: Optional[float] = None
     min_cold_load: Optional[float] = None
-    target_util: Optional[float] = None
-    cold_mult: Optional[float] = None
-    cold_workers: Optional[int] = None
     max_workers: Optional[int] = None
-    test_workers: Optional[int] = None
```
LucasArmandVast, line 27: I think we don't want to get rid of this... since the API will still return these values even though they don't mean anything... worth confirming though.

madisonvast: Confirmed, and I think we're safe here. WorkergroupConfig is a write-only model: to_dict() only, no from_dict, and all 8 call sites construct it to POST. The read path (show_workergroups) returns raw dicts from rj["results"] and never touches the dataclass, so everything the API returns still reaches the user unchanged, I verified with a stubbed response carrying all four fields.

The clincher is that WorkergroupConfig(**api_response) was already a TypeError before this PR: the response carries id, api_key, created_at, user_id, and search_query (vs the config's search_params), none of which the old class accepted either. So it was never usable for parsing responses in the first place.

Compare vastai/data/endpoint.py, which has both halves: EndpointConfig for writes and EndpointData.from_dict() for reads. Workergroups only ever had the write half. If we do want a typed read model later, the right shape is a WorkergroupData.from_dict() mirroring EndpointData — that one would carry the dead fields as returned-but-meaningless, plus id/api_key/user_id. Happy to file that separately if it's useful.

(Side note: test_workers isn't in the GET response at all, only min_load/target_util/cold_mult/cold_workers come back.)

LucasArmandVast (approved):

